PR #1, opened 2018-10-19
Add first job template & script for environment installation (without…

tremes:
… bastion)

### DIFF
--- a/scripts/install-integr8ly-no-bastion.groovy
+++ b/scripts/install-integr8ly-no-bastion.groovy
@@ -1,0 +1,27 @@
+
+node("${NODE}") {
+        deleteDir()
+        stage('Checkout'){
+            checkout scm: [
+                $class: "GitSCM", 
+                branches:[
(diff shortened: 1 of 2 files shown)


scripts/install-integr8ly-no-bastion.groovy
```diff
@@ -0,0 +1,27 @@
...
+                     name: "origin", 
+                     refspec: "+refs/heads/*:refs/remotes/origin/*",
+                     url: "https://github.com/${OWNER}/installation.git" ]
+                   ]
+                ]
+        }
+        
+        stage('Install'){
+            sh '''
+                sed -i 's/rhsso_seed_users_count: 50/rhsso_seed_users_count: 2/g' evals/roles/rhsso/defaults/main.yml
```
tremes: This is just to limit number of users and it's optional

tremes:
@trepel Please take a look.

PR #4, opened 2018-10-26
Add installation pipeline template

trepel:
*Motivation*

To have the pipeline configuration versioned and to have an easy way to create the pipeline on different jenkins instances. The pipeline according to this definition is already created here:
https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/installation-pipeline/

*To Test*
execute this:

```
jenkins-jobs --conf jenkins_jobs.ini test jobs/templates/installation-pipeline.yaml
```

Optionally you can also execute 'update' and run against QE Pony cluster, input parameters can be found here:
https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/installation-pipeline-bastion/18/parameters/

### DIFF
--- a/jobs/templates/installation-pipeline.yaml
+++ b/jobs/templates/installation-pipeline.yaml
@@ -1,0 +1,150 @@
+---
+
+- job:
+    name: installation-pipeline
+    project-type: pipeline
+    description: "Installs Integreatly by using bastion server as Jenkins lave and executing Ansible installer there."
+    sandbox: false
+    parameters:
+      - string:
+          name: REPOSITORY
+          default: https://github.com/integr8ly/installation.git
+          description: "Repository of the Integreatly installer"
+      - string:
+          name: BRANCH
+          default: 'master'
+          description: "Branch of the installer repository"
+      - string:
+          name: GH_CLIENT_ID
+          description: "GitHub Client ID for OAuth Apps, required for some of the walkthroughs. Can be left empty"
+      - string:
+          name: GH_CLIENT_SECRET
+          description: "GitHub Client Secret for OAuth Apps, required for some of the walkthroughs. Can be left empty"
+      - bool:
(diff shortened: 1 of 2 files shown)


trepel:
@tremes could you please have a look - for some reason I cannot assign you for review

jobs/templates/installation-pipeline.yaml
```diff
@@ -0,0 +1,150 @@
...
+      import jenkins.model.*
+      import hudson.slaves.*
+      import hudson.slaves.EnvironmentVariablesNodeProperty.Entry
+
+      import hudson.plugins.sshslaves.verifiers.*
+
+      String bastionLabel = "${BASTION_URL}-slave"
+
+      timeout(60) {
+          node('cirhos_rhel7') {
```
tremes: Shouldn't we parametrize this as well? WDYT?

tremes:
@trepel Added one minor comment. Otherwise it looks good. Well I am not a member of Integr8ly group so  that's the reason why I can't be assigned for a review.

jobs/templates/installation-pipeline.yaml
tremes: Well we can have differrent machines and labels than only `cirhos_rhel7` or not?

trepel: Yes, but such slaves will likely serve for specific purposes only. For instance we have various bastion slaves but such slaves you want only use for the pipeline that deals with specific cluster behind that specific bastion. Using 'cirhos_rhel7' is nice limitation to any non-specific slave.

Can you think of any specific use case when one would want to run this pipeline on specific slave?

tremes: Well I think I can't...I just hope nobody will change our jenkins labels or something like that.:-)

PR #7, opened 2018-10-30
Add template for managed-service-broker tests pipeline.

tremes:


jobs/templates/managed-service-broker-tests.yaml
```diff
@@ -0,0 +1,20 @@
...
+            url: http://github.com/integr8ly/ci-cd
+            clean: true
+            branches: 
+              - master
+      script-path: scripts/msb-tests.groovy
+      lightweight-checkout: true
+    parameters:
+      - string:
+          name: CLUSTER_ROUTE
+          default: 'integr8ly-staging.dynamic.xpaas'
```
trepel: Could you please add some description as well ?

trepel: For both params if possible

tremes:
@trepel Please take a look...and possibly merge.

scripts/msb-tests.groovy
```diff
@@ -0,0 +1,63 @@
...
+      deleteDir()
+
+      withEnv(["GOPATH=$WORKSPACE",
+               "BROKER_URL=http://msb-${NAMESPACE}.apps.${CLUSTER_ROUTE}"
+            ])
+            {
+                  stage('Clone the managed-service-broker repo'){
+                        dir(".") {
+                              git branch: 'master', url: 'https://github.com/integr8ly/managed-service-broker.git'
+                              if(!fileExists(file: './Makefile')) error("Error cloning managed-service-broker repo")
```
trepel: I know I used to use this too but I think it is not required. In case of clone failure the `git` command fails with error, no need for explicit check afterwards.

scripts/msb-tests.groovy
```diff
@@ -0,0 +1,63 @@
...
+                        dir("$GOPATH/src/github.com/integr8ly"){
+                              sh '''
+                                    ln -sf $WORKSPACE managed-service-broker
+                                    cd managed-service-broker
+            
+                                    make build_binary 
+                                    sudo docker login -u ${QUAY_USERNAME} -p ${QUAY_PASSWORD} quay.io
+                                    sudo docker build -t quay.io/integr8lyqe/managed-services-broker:1.0.0 -f ./tmp/build/broker/Dockerfile .
+                                    sudo docker push quay.io/integr8lyqe/managed-services-broker:1.0.0
+                              '''
```
trepel: Would it be possible to use different tag? `1.0.0` seems as if it is stable version but it is just what is currently in master branch

scripts/msb-tests.groovy
tremes: Yeah..copy/paste from you :-)

scripts/msb-tests.groovy
tremes: which tag do you suggest? latest?:)

jobs/templates/managed-service-broker-tests.yaml
tremes: OK

scripts/msb-tests.groovy
trepel: Yes, `latest` sound good. Other options are `master` or even something like `0.0.1`. I leave it up to you, we can stick with `latest`, that's fine.

PR #11, opened 2018-11-07
Simplify browser testsuite pipeline

tremes:


tremes:
@trepel Please take a look.

PR #16, opened 2018-12-04
Add release create and helper jobs

mikenairn:
## Motivation

Automate the process of creating new **minor** releases of integrealy.

## What

Add release create job
Add delete docker tag and repo branches and tags jobs
Add release and repos views
Add generate inline and configure_jenkins scripts

## Related:

* https://github.com/integr8ly/installation/pull/228

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,3 +30,13 @@ jenkins-jobs --conf /path/to/your/jenkins_jobs.ini test /path/to/your/template.y
 ```
 jenkins-jobs --conf /path/to/your/jenkins_jobs.ini update /path/to/your/template.yaml
 ```
+
+### Generate inline script jobs
+```
+./scripts/generate_inline_script_pipeline_job -j /path/to/your/template.yaml -o /path/to/your/generated/jobs
+```
+
+### Configure all jobs and views
+```
+./scripts/configure_jenkins /path/to/your/jenkins_jobs.ini
+```
--- a/jobs/release/release-create/Jenkinsfile
+++ b/jobs/release/release-create/Jenkinsfile
@@ -1,0 +1,209 @@
+#!groovy
+
+//Helper methods
+def checkoutGitRepo(gitUrl, gitRef, credentialsID) {
+    checkout([$class: 'GitSCM', branches: [[name: gitRef]],
+            doGenerateSubmoduleConfigurations: false,
+            extensions: [],
+            submoduleCfg: [],
+            userRemoteConfigs: [[credentialsId: credentialsID, url: gitUrl]]])
+}
+
+String gitRepoIsDirty(untrackedFiles = 'no') {
+    return sh(
+            returnStdout: true,
+            script: "git status --porcelain --untracked-files=${untrackedFiles}"
+    )?.trim()
+}
+
+String getReleaseBranch(version) {
+    def versionParts = version.tokenize(".")
+    return "v${versionParts[0]}.${versionParts[1]}"
+}
+
+String getReleaseTag(version, candidate = null) {
+    if (candidate) {
+        "release-${version}-${candidate}"
+    } else {
+        "release-${version}"
+    }
+}
+//
+
+def installationGitUrl = params.installationGitUrl
+def installationGitRef = params.installationGitRef
+
+def releaseBaseBranch = 'master'
+def releaseBranchName = getReleaseBranch(params.releaseVersion)
+def releaseTagName = getReleaseTag(params.releaseVersion)
+
+def releaseConfig = [:]
+def componentConfigs = []
+
+def githubCredentialsID = "jenkinsgithub"
+def quayCredentialsID = "integreatly-quay"
+
+currentBuild.displayName = "${currentBuild.displayName} ${params.releaseVersion}"
+
+node('cirhos_rhel7') {
+    sshagent([githubCredentialsID]) {
+        sh "git config --global user.name  \"Jenkins\""
+        sh "git config --global user.email \"jenkins@wendy.feedhenry.net\""
+
+        cleanWs()
+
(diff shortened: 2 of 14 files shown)


jobs/release/release-create/Jenkinsfile
```diff
@@ -0,0 +1,199 @@
...
+                    sh "git push origin HEAD:${releaseBranchName}"
+                }
+
+                sh "git tag ${releaseTagName}"
+                sh "git push origin ${releaseTagName}"
+            }
+        }
+
+    }
+}
```
mikenairn: Hope the stages here are fairly self explanatory, but this is the basic process:

1. Pull Installation repo and check that we don't already have the release tag we are about to try and create, exit if it does.
1. Fetch all component repos as described in the components.yaml file (https://github.com/integr8ly/installation/blob/a9026074c319ecae034858c0255151bb7b71602b/COMPONENTS.yaml)
1. Create a release branch of vX.X (e.g v1.2) on the installation repo **only**. Always currently created from master as per a new minor release.
1. Create a release tag on each components git repo e.g release.1.2.0 at the HEAD of the master branch at that time
1. Create a docker image tag in the components quay repo which points the current HEAD short hash tag at a new release tag e.g 1234567 -> release-1.2.0. (Note, it is expected that something will have already created a tag in the components quay repo with a tag of the current HEADs short commit hash see https://github.com/integr8ly/3scale-operator/blob/master/.travis.yml#L29)
1. Update docker image versions and git refs as appropriate for each component in the installation repos release branch, and tag with the same release e.g. release-1.2.0

scripts/configure_jenkins.sh
```diff
@@ -0,0 +1,27 @@
...
+generate_inline_script_job $SCRIPTS_DIR/../jobs/release/release-delete/integreatly-release-delete.yaml
+generate_inline_script_job $SCRIPTS_DIR/../jobs/repos/repos-delete-branches-and-tags/repos-delete-branches-and-tags.yaml
+generate_inline_script_job $SCRIPTS_DIR/../jobs/repos/repos-delete-docker-image-tags/repos-delete-docker-image-tags.yaml
+
+#Generated jobs
+jenkins-jobs --conf $CONFIG update $SCRIPTS_DIR/../jobs/generated/
+
+#Views
+jenkins-jobs --conf $CONFIG update $SCRIPTS_DIR/../views/repos/
+jenkins-jobs --conf $CONFIG update $SCRIPTS_DIR/../views/release/
```
mikenairn: You can use this script to just configure a given jenkins instance with all the jobs:

```
./scripts/configure_jenkins.sh myjenkins.ini
```

jobs/release/release-create/Jenkinsfile
mikenairn: @maleck13 FYI, you happy enough with this? ^

trepel:
Reviewing

README.md
```diff
@@ -30,3 +30,13 @@ jenkins-jobs --conf /path/to/your/jenkins_jobs.ini test /path/to/your/template.y
...
 ```
+
+### Generate inline script jobs
+```
+./scripts/generate_inline_script_pipeline_job -j /path/to/your/template.yaml -o /path/to/your/generated/jobs
+```
+
+### Configure all jobs and views
+```
+./scripts/configure_jenkins /path/to/your/jenkins_jobs.ini
```
trepel: This didn't work for me when `jenkins-job.ini` was not located in `ci-cd` directory or its subdirectories. Using both relative and absolute path to e.g. `ci-cd/jenkins_jobs.ini` worked as expected. Not a blocker for a merge though.

tremes:
This all looks good to me, but I think this doesn't give any room for testing a release (something like "staging" stage). Wouldn't be helpful to split it up to two phases? First create a branch with all updated components, then there would be QE testing and then create a release tag. I might be wrong so WDYT? 

I know this would be more complicated, because it would likely require creating release branches for each component, but on the other hand what we will do when we realize there is some bug after release? Will we make another release in such case? If so then it's really not a problem....

jobs/release/release-create/Jenkinsfile
```diff
@@ -0,0 +1,199 @@
...
+
+String getReleaseTag(version, candidate = null) {
+    if (candidate) {
+        "release-${version}-${candidate}"
+    } else {
+        "release-${version}"
+    }
+}
+//
```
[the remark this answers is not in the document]
mikenairn: Yes, that's how we have done it in the past to avoid lots of duplication.

jobs/release/release-create/Jenkinsfile
mikenairn: Since creating this pipeline we have discussed and started looking into a slightly different way of releasing where the integreatly components won;t be treated any differently from external project components. This will mean this flow and pipeline will end up changing considerably.

We can keep it this way for now though so if the need arises before that work is complete we can at least create a new release.

README.md
mikenairn: Hadn't even tried that tbh, but yeah the script is assuming you are working with a file in this directory or a sub directory of it as you have pointed out.

mikenairn:
> This all looks good to me, but I think this doesn't give any room for testing a release (something like "staging" stage). Wouldn't be helpful to split it up to two phases? First create a branch with all updated components, then there would be QE testing and then create a release tag. I might be wrong so WDYT?

@tremes We could certainly have the pipeline pause between stages to allow for some testing before continuing if needs be.

I'd consider a first cut though, we can iterate on this to add any improvements. Also this https://github.com/integr8ly/ci-cd/pull/16#discussion_r240522923 will have an impact on this pipeline, so process improvements can made as part of that.

tremes:
@mikenairn OK I agree.

PR #18, opened 2018-12-10
browser-based-pipeline set build result as unstable when there's a te…

tremes:
…st failure

## Motivation
Mark a build result as unstable instead of failure when there is some test failure

## What
Add try/catch block for "npm test" command and set build result in the catch block

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/scripts/browser-tests.groovy
+++ b/scripts/browser-tests.groovy
@@ -13,8 +13,12 @@ timeout(60) {
                       # Disable starting Selenium, Docker is used instead
                       sed -i 's/"start_process" : true,/"start_process" : false,/g' nightwatch.json
                       npm install
-                      npm test
                     """
+                    try {
+                        sh "npm test"
+                    } catch (exp) {
+                        currentBuild.result = 'UNSTABLE'
+                    }
                 } 
             }
         }


scripts/browser-tests.groovy
```diff
@@ -13,8 +13,12 @@ timeout(60) {
                       # Disable starting Selenium, Docker is used instead
                       sed -i 's/"start_process" : true,/"start_process" : false,/g' nightwatch.json
                       npm install
-                      npm test
                     """
+                    try {
+                        sh "npm test"
+                    } catch (exp) {
+                        currentBuild.result = 'UNSTABLE'
```
trepel: If you do just this, the exception stack trace and message is not shown in the build log. I think it is pretty important to see the full stack trace in case of error.

I tried using `exp.printStackTrace()` as the very first command in `catch` block but it didn't work. Not sure how to overcome this. But seeing the stack trace seems more important to me than 'FAILURE' vs 'UNSTABLE' result.

tremes: Ahh OK. Hm so maybe `println(exp.getStackTrace())`. Can give it a try.

trepel: Nope, I can see only `[Ljava.lang.StackTraceElement;@2404307f` in the log

tremes:
I think it's not so helpful as I hoped...

PR #31, opened 2019-01-24
created installation-smoke-tests.yaml jjb

pawelpaszki:
## Motivation
https://issues.jboss.org/browse/INTLY-561

## What
Create jjb for clean-uninstallation-pipeline

## Verification Steps
Check is to run this command inside the repository with PR changes to confirm that the update is successful (and also make sure that the pipeline configuration does not change -> as the template is the same as definition on Jenkins): 

`jenkins-jobs --conf jenkins_jobs.ini update jobs/integr8ly/installation-smoke-tests.yaml`

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO


pawelpaszki:
@matskiv @tremes @trepel creating PR now. Please don't verify before we resolve the issue with our Jenkins slaves

pawelpaszki:
@matskiv @trepel could you please have a look now? The successful build was executed here: https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/installation-smoke-tests/189/console

jobs/integr8ly/installation-smoke-tests.yaml
```diff
@@ -37,4 +28,27 @@
...
+                        npm install
+                    '''
+                }
+                dir('tests/backend-testsuite'){    
+                    try {
+                        sh '''
+                            gulp smoke
+                        '''
+                    } catch(exp) {
+                        currentBuild.result = 'UNSTABLE'
```
matskiv: btw, @tremes @trepel do you guys know what would be the result of the build if we didn't catch failure here and just let sh/gulp exit with non-zero code?

tremes: The result is FAILURE in such case

matskiv: Why do we prefer UNSTABLE instead of FAILURE?

tremes: No strong opinion here. Would you prefer FAILURE?

matskiv: I don't have a strong opinion about the state. But wrapping it in try/catch seemed useless to me at first.

P.S: This is kinda of topic for this PR, sorry for starting this conversation.

tremes: Yes this was kinda my experiment TBH...and you're probably right that's useless (just yellow color instead of red...from time to time:-))

pawelpaszki: @matskiv @tremes any suggestions about the try/catch, so the PR can be merged?

pawelpaszki: removing try/catch block in such case and merging

PR #32, opened 2019-01-28
release monitoring job

odra:
Adds releasing monitoring for 3scale repo.

## Verification steps:

Add the job in jenkins:

```
jenkins-jobs --conf  jenkins_job.ini update jobs/release-monitoring/release-monitoring-3scale/release-monitoring-3scale.yaml
```

Trigger the job (needs a github api token with repo access enabled).

It should create a new "next" repo (3scale-next in this case) with the new identified version or abort the job run in case no new version is available.

I will change the job definition to use the installer upstream repo once the pr is in good shape to be merged.

### DIFF
--- a/jobs/release-monitoring/release-monitoring-github/Jenkinsfile
+++ b/jobs/release-monitoring/release-monitoring-github/Jenkinsfile
@@ -1,0 +1,201 @@
+#!groovy
+
+//remote repository checkout
+def checkoutGitRepo(gitUrl, gitRef, credentialsID) {
+    checkout([$class: 'GitSCM', branches: [[name: gitRef]],
+            doGenerateSubmoduleConfigurations: false,
+            extensions: [],
+            submoduleCfg: [],
+            userRemoteConfigs: [[credentialsId: credentialsID, url: gitUrl]]])
+}
+
+//gets latest release tag of a github repository
+def getLatestRelease(gitOrg, gitRepo, gitTokenId) {
+    def response = null
+    def data = null
+    
+    withCredentials([string(credentialsId: gitTokenId, variable: 'gitToken')]) {
+        def url = "https://api.github.com/repos/${gitOrg}/${gitRepo}/releases/latest"
+        def headers = [
+            ["name": "Authorization", "value": "token ${env.gitToken}"]
+        ]
+
+        response = httpRequest customHeaders: headers, url: url
+        data = readJSON text: response.content
+    }
+
+    return [response.status, data]
+}
+
+def getLabelScore(label) {
+    if (!label) {
+        return 0
+    }
+    
+    def score = 0
+    def scores = [
+        'GA': 1,
+        'redhat': 1
+    ]
+    def parts = label.tokenize('-')
+    
+    parts.each { part ->
+        if (scores.containsKey(part)) {
+            score += scores[part]
+        }
+
+        if (!scores.containsKey(part) && part.isInteger()) {
+            score += part.toInteger()
+        }
+    }
+
+    return score
+}
+
+//checks if current semver is lower than the supposed latest one
+def hasNewGARelease(currentVersion, newVersion) {
+    def current = currentVersion.tokenize('.')
+    def latest = newVersion.tokenize('.')
+
+    def currentMajor = current[0] as Integer
+    def latestMajor = latest[0] as Integer
+    def currentMinor = current[1] as Integer
+    def latestMinor = latest[1] as Integer
+    def currentPatch = current[2] as Integer
+    def latestPatch = latest[2] as Integer
+    def currentLabel = current[3]
+    def latestLabel = latest[3]
+
+    def previousVer = 0
+    def currentVer = 0
+    for (i = 0; i <= 2; i++) {
+        if (i > 0) {
+            previousVer = i - 1
+        }
+        currentVer = i
+
+        def previousDiff = current[previousVer] == latest[previousVer]
+        if (i == 0) {
+            previousDiff = true
+        }
+        def currentDiff =  latest[currentVer] > current[currentVer]
+        if (previousDiff && currentDiff) {
+            return true
+        }
+    }
+
+    def currentLabelScore = getLabelScore(currentLabel)
+    def latestLabelScore = getLabelScore(latestLabel)
+
+    if (latestLabelScore > currentLabelScore) {
+        return true
+    }
+
+    return false
+}
+
+def fnMap = [
+    '3scale': { currentVersion, newVersion -> hasNewGARelease(currentVersion, newVersion) }
+]
+
+def installationGitUrl = params.installationGitUrl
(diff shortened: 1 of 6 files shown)


maleck13:
Is this something we actually want to merge to do we want to wait until we have tried the Msg bus version first

odra:
I would merge it (when tested) so we can refine the install -> smoke test -> uninstall workflow

jobs/release-monitoring/release-monitoring-3scale/Jenkinsfile
```diff
@@ -0,0 +1,202 @@
...
+        releaseConfig[manifestVar] = latestRelease
+        dir('installation') {
+            sshagent([githubCredentialsID]) {
+                sh 'git config --global user.name "Automatron"'
+                sh 'git config --global user.email "github.eng@feedhenry.com"'
+                def branchCode = sh script: "git checkout -b ${nextBranch}", returnStatus: true
+                if (branchCode  != 0) {
+                    sh "git checkout ${nextBranch}"
+                }
+                sh "git fetch origin && git reset --hard origin/${nextBranch} || true"
```
mikenairn: What's the point of this?

jobs/release-monitoring/release-monitoring-3scale/Jenkinsfile
```diff
@@ -0,0 +1,202 @@
...
+                string(name: 'ANSIBLE_USER', value: clusterConfig['ANSIBLE_USER']),
+                string(name: 'MASTER_URLS', value: clusterConfig['MASTER_URLS']),
+                string(name: 'BASTION_USER', value: clusterConfig['BASTION_USER']),
+                string(name: 'BASTION_URL', value: clusterConfig['BASTION_URL']),
+                string(name: 'BASTION_PRIVATE_KEY_ID', value: clusterConfig['BASTION_PRIVATE_KEY_ID'])
+            ]
+
+            build job: uninstallJobName, parameters: uninstallJobParams   
+        }
+    }
```
mikenairn: I think we will eventually trigger a job that does all this rather than doing each part explicitly in here. @trepel Is apparently working on a pipeline that does this based on jobs that already exist and we should re-use those where possible.

jobs/release-monitoring/release-monitoring-3scale/release-monitoring-3scale.yaml
```diff
@@ -0,0 +1,73 @@
...
+      
+    pipeline-scm:
+      script-path: jobs/release-monitoring/release-monitoring-3scale/Jenkinsfile
+      scm:
+        - git:
+            branches:
+              #- master
+              - 'release-monitoring'
+            #url: 'https://github.com/integr8ly/ci-cd.git'
+            url: 'https://github.com/odra/ci-cd.git'
```
mikenairn: Needs changed to whatever it's supposed to be. You should also add a line into the configure_jenksin script to have it generated and pushed to the Jenkins instance https://github.com/integr8ly/ci-cd/blob/master/scripts/configure_jenkins.sh#L20

jobs/release-monitoring/release-monitoring-3scale/Jenkinsfile
odra: Agreed, I did it for the sake of the poc but I was wondering if we should had a nightly job for those "next" branches instead of triggering it on each product release

jobs/release-monitoring/release-monitoring-3scale/release-monitoring-3scale.yaml
odra: Using it this way to test the changes and I will revert it back to master once the pr is in good shape

jobs/release-monitoring/release-monitoring-3scale/Jenkinsfile
odra: to reset to the latest next branch commit if we already have one (in case we have more than one product release before merging it back to master)

jobs/release-monitoring/release-monitoring-3scale/Jenkinsfile
mikenairn: Yeah. I know we discussed this a bit, but thinking about it more, it makes sense to me to separate these things out as much as possible. I think a job that runs nightly that just checks for branches matching a given pattern and runs the appropriate install/test/unsintall pipeline make sense.

jobs/release-monitoring/release-monitoring-3scale/release-monitoring-3scale.yaml
mikenairn: Using the script thing i sent you it doesn't load from SCM, so you can make change locally and test it more easily that way.

jobs/release-monitoring/release-monitoring-3scale/Jenkinsfile
```diff
@@ -0,0 +1,202 @@
...
+def testJobName = params.testJobName
+def uninstallJobName = params.uninstallJobName
+
+node {
+    stage('Fetch Installation Repo') {
+        println("Fetch Installation Repo")
+
+        cleanWs()
+        dir('installation') {
+            checkoutGitRepo(installationGitUrl, installationGitRef, githubCredentialsID)
```
mikenairn: Why do we pass the installationGitRef in here? What is it supposed to be set to? I'd imagine it should always be master no?

jobs/release-monitoring/release-monitoring-3scale/Jenkinsfile
mikenairn: OK, so you could do it in the if  above? and expect it to work? Should you not also be rebasing master into this branch at some point here?

odra:
@tremes @trepel  could you please take a look at this pr

tremes:
@odra sure I'll take a look today

odra:
@mikenairn I've updated the pr with a couple of changes, mind reviewing it again?

jobs/release-monitoring/release-monitoring-3scale/release-monitoring-3scale.yaml
```diff
@@ -0,0 +1,50 @@
...
+          description: '[REQUIRED] github ssh credential id to be used'
+    pipeline-scm:
+      script-path: jobs/release-monitoring/release-monitoring-3scale/Jenkinsfile
+      scm:
+        - git:
+            branches:
+              - master
+            url: 'https://github.com/integr8ly/ci-cd.git'
+            skip-tag: true
+            wipe-workspace: false
```
mikenairn: Since this is fully paramateriased and all these project release discovery jobs will be exactly the same, apart from the tag processing part, would it not make more sense to make this a generic github-release-monitoring job and create another that wraps this and passes the 3scale params into it? That's what we planned to do with the other pipelines anyway so would be more consistent and reduce the duplication.

PR #35, opened 2019-01-30
Smoke tests UNSTABLE on test errors/failures

trepel:
## What & Why & Motivation
Smoke test job build result should be UNSTABLE in case of test errors/failures. Not failure.

## How
shell script that runs the tests always exits with 0. Outup of the script is parsed to see whether there were test failures or errors. Based on that the build result is set. In case of error the publishing is skipped (no xml report is generated in that case anyway)

## Verification Steps
* `jenkins-jobs --conf jenkins_jobs.ini test jobs/integr8ly/installation-smoke-tests.yaml`
* Run the test suite - you should see 3scale operator failure since this has been removed yesterday. Should be reported, build should be UNSTABLE
* Remove the failing test - should be reported and build should be SUCCESS
* Do some syntax error in gulpfile - should not be reported and build should be FAILURE

Not sure if is is really required to do all this on Jenkins, the change is relatively simple.

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/jobs/integr8ly/installation-smoke-tests.yaml
+++ b/jobs/integr8ly/installation-smoke-tests.yaml
@@ -30,6 +30,9 @@
           description: 'Provide release version when testing particular release.'
     dsl: |
         node('cirhos_rhel7'){
+
+            Boolean publishTestResults = true
+
             stage('Clone QE repository') {
                 dir('.') {
                     git branch: "${BRANCH}", url: "${REPOSITORY}"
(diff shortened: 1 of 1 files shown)


trepel:
If we agree on this, we can do similar modifications for other pipelines

tremes:
Hm but does it actually work? https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/installation-smoke-tests/214/console

trepel:
> Hm but does it actually work? https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/installation-smoke-tests/214/console

Yes, it does. But great point, it does not print the output to the console anymore. It is only visible in the test report. I added `println output` so that it gets to console log as well

trepel:
@pawelpaszki @tremes Ah, you are so super quick with reviews, I really need to learn this from you. Thanks. Merging.

PR #36, opened 2019-01-30
Add pipelines for install/uninstall for the new QE Pony

tremes:
## Motivation
QE pony cluster has been reinstalled and it doesn't use a bastion server. This requires a little bit different pipelines for install/uninstall.

### DIFF
--- a/jobs/integr8ly/installation-pipeline-qe-pony.yaml
+++ b/jobs/integr8ly/installation-pipeline-qe-pony.yaml
@@ -1,0 +1,104 @@
+---
+
+- job:
+    name: installation-pipeline-qe-pony
+    project-type: pipeline
+    description: "Installs Integreatly remotely from Jenkins slave to QE Pony cluster."
+    sandbox: false
+    concurrent: false
+    parameters:
+        - string:
+            name: REPOSITORY
+            default: https://github.com/integr8ly/installation.git
+            description: "Repository of the Integreatly installer"
+        - string:
+            name: BRANCH
+            default: 'master'
+            description: "Branch of the installer repository"
+        - string:
+            name: CLUSTER_URL
+            description: "URL of cluster for installation."
+        - string:
+            name: GH_CLIENT_ID
+            default: b15294d65592a5e6afb4
+            description: "GitHub Client ID for OAuth Apps, required for some of the walkthroughs. Can be left empty"
+        - string:
+            name: GH_CLIENT_SECRET
+            default: 539a4a94cda8854d610d2dbef25881a31b49cba2
+            description: "GitHub Client Secret for OAuth Apps, required for some of the walkthroughs. Can be left empty"
+        - bool:
+            name: SELF_SIGNED_CERTS
+            default: false
+            description: "Indicates whether cluster uses self signed certificates or not"
+        - string:
+            name: ANSIBLE_USER
+            default: root
+            description: "User for Ansible to access the master node of target cluster"
+        - string:
+            name: MASTER_URLS
+            description: "Comma separated list of URLs for master nodes of target cluster to be used in Ansible inventory file"
+        - string:
+            name: OC_USER
+            description: "OpenShift user for QE cluster."
+        - string:
+            name: OC_PASSWORD
+            description: "OpenShift password for QE cluster."
+    dsl: |
+        timeout(60) { ansiColor('gnome-terminal') { timestamps {
+            node('cirhos_rhel7') {        
+                stage('Verify input') {
+                    if (!MASTER_URLS) {
+                        throw new hudson.AbortException('MASTER_URLS parameter is required!')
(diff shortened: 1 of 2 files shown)


tremes:
Please review @trepel, @pawelpaszki and @matskiv

jobs/integr8ly/installation-pipeline-qe-pony.yaml
```diff
@@ -0,0 +1,93 @@
...
+            name: OC_USER
+            description: "OpenShift user for QE cluster."
+        - string:
+            name: OC_PASSWORD
+            description: "OpenShift password for QE cluster."
+    dsl: |
+        timeout(60) {
+            node('cirhos_rhel7') {        
+                stage('Verify input') {
+                    if (!MASTER_URLS) {
```
pawelpaszki: Why there is only MASTER_URLS missing check, if there are some other required parameters. Maybe we should check for those as well if we wanna do it for MASTER_URLS

jobs/integr8ly/uninstallation-pipeline-qe-pony.yaml
```diff
@@ -0,0 +1,73 @@
...
+                    dir('installation') {
+                        git branch: BRANCH, url: REPOSITORY
+                    } // dir
+                } // stage
+
+                stage('Prepare environment') {
+                    dir('installation') {
+                        sh """
+                            cp ./evals/inventories/hosts.template ./evals/inventories/hosts
+                            sed -i 's/ansible_user=ec2-user/ansible_user=${ANSIBLE_USER}/g' ./evals/inventories/hosts
```
pawelpaszki: I don't see ANSIBLE_USER parameter specified in the list of parameters

tremes: Oh yeah good catch. THanks

jobs/integr8ly/uninstallation-pipeline-qe-pony.yaml
```diff
@@ -0,0 +1,73 @@
...
+            name: OC_USER
+            description: "OpenShift user for QE cluster."
+        - string:
+            name: OC_PASSWORD
+            description: "OpenShift password for QE cluster."
+    dsl: |
+        timeout(60) {
+            node('cirhos_rhel7') {        
+                stage('Verify input') {
+                    if (!MASTER_URLS) {
```
pawelpaszki: the same as above

jobs/integr8ly/installation-pipeline-qe-pony.yaml
tremes: Well yes it would make sense to check CLUSTER URL and USER credentials.

jobs/integr8ly/installation-pipeline-qe-pony.yaml
```diff
@@ -0,0 +1,102 @@
...
+            description: "Branch of the installer repository"
+        - string:
+            name: CLUSTER_URL
+            description: "URL of cluster for installation."
+        - string:
+            name: GH_CLIENT_ID
+            description: "GitHub Client ID for OAuth Apps, required for some of the walkthroughs. Can be left empty"
+        - string:
+            name: GH_CLIENT_SECRET
+            description: "GitHub Client Secret for OAuth Apps, required for some of the walkthroughs. Can be left empty"
```
matskiv: Improvement idea: somebody could create GH App just for Pony and we could have it as default value for these params, or hard-coded all together.

trepel: There can be used (hard-coded as defaults):
b15294d65592a5e6afb4
539a4a94cda8854d610d2dbef25881a31b49cba2

tremes: Yeah why not...I'll update

jobs/integr8ly/installation-pipeline-qe-pony.yaml
```diff
@@ -0,0 +1,104 @@
+---
+
+- job:
+    name: installation-pipeline-qe-pony
+    project-type: pipeline
+    description: "Installs Integreatly remotely from Jenkins slave to QE Pony cluster."
+    sandbox: false
+    concurrent: true
```
trepel: Since this is qe pony specific, do we want to allow concurrent builds? I would say there is no point in running install playbook twice against the same cluster in parallel.

tremes: Yes let's remove this

tremes:
I think I should add the fancy time displaying plugin to these

pawelpaszki:
yeah. that makes the console output much nicer

tremes:
Updated. I think it's ok to merge. Please review.

PR #37, opened 2019-01-30
changed pipeline to generate reports despite of test failures

pawelpaszki:
## What & Why & Motivation
After-first-login-tests job build result should be UNSTABLE in case of test errors/failures. Not failure.

## How
shell script that runs the tests always exits with 0. Output of the script is parsed to see whether there were test failures or errors. Based on that the build result is set. In case of error the publishing is skipped (no xml report is generated in that case anyway)

## Verification Steps
Go to this pipeline (with the definition the same as the target pipeline): https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/ppaszki-after-first-login/ and check the output of builds 3. and 4. to confirm that the reports are generated with and without failures

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/jobs/integr8ly/after-first-login-tests.yaml
+++ b/jobs/integr8ly/after-first-login-tests.yaml
@@ -31,6 +31,9 @@
           description: 'This optional parameter allows to test particular release version. If not provided, master will be used'
     dsl: |
         node('cirhos_rhel7') {
+            
+            Boolean publishTestResults = true
+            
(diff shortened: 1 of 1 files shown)


trepel:
Tried your job on my cluster (it is being provisioned at the moment). It failed (expected) and the report correctly showed the failure.

PR #39, opened 2019-01-31
changed pipeline to generate reports despite of test failures

pawelpaszki:
## What & Why & Motivation
browser-based-testsuite-pipeline job build result should be UNSTABLE in case of test errors/failures. Not failure.

## How
shell script that runs the tests always exits with 0. Output of the script is parsed to see whether there were test failures or errors. Based on that the build result is set. In case of error the publishing is skipped (no xml report is generated in that case anyway)

## Verification Steps
Go to this pipeline (with the definition the same as the target pipeline): https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/ppaszki-browser-based/7/console and check the output to confirm that the results are published

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO


jobs/integr8ly/browser-based-testsuite-pipeline.yaml
```diff
@@ -48,17 +52,32 @@
...
                             """
+                            
+                            String output = sh returnStdout: true, script: 'npm test || true'
+                            
+                            // To see the output in Jenkins build console log as well
+                            println output
+
+                            if(!output.contains('Enmasse Console Login')) {
+                                currentBuild.result = 'FAILURE'
+                                publishTestResults = false
```
matskiv: What is the purpose of this condition branch?

pawelpaszki: to check if the tests were executed

matskiv: Do we care if they were not executed or they failed?
I just think we could simplify this by not caring.
E.g:
```
try {
  sh 'npm test'
} catch (Exception e) {
  currentBuild.result = 'FAILURE'
}
```
With this^ we won't need any conditions here, nor in the reporting.

pawelpaszki: need to try it out. Will create some broken branch and see if the report is generated. Can change it then

trepel: @matskiv the `npm tests` exits with 1 in case of test failure. I wanted to ensure that the build is marked as UNSTABLE for test failures and as FAILURE for errors during running tests. In case of such errors the report is not generated at all, so I thought I would skip the reporting stage.

matskiv: @trepel but the solution that achieves that isn't very elegant IMHO. But this are just my 2cents, feel free to ignore.

trepel: @matskiv no, it is not indeed. It is just a solution I was able to come up with. I don't want to ignore, I would be interested in how to do it better.

matskiv: One alternative that comes to my mind: in the catch block, check if there are any files in reports directory - if yes = unstable, if no = failure.

pawelpaszki: @matskiv I was thinking about your latest suggestion. This would introduce some more code and as a result would not be much cleaner solution. You would need to get the number of entries in the **reports** directory (`ls | wc -l`), then trim the output (unfortunately the output is padded with empty spaces) and check if it is not equal to 0. Do you think it is worth the hassle?

matskiv: I think that's cleaner than stealing stdout, printing it and checking `output.contains('Enmasse Console Login')`

pawelpaszki: OK. I will do it this way. You are probably right that using wc ... is easier to understand

pawelpaszki:
@matskiv can you please have another look? Here is a build without failures: https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/ppaszki-browser-based/13/console and with failure: https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/ppaszki-browser-based/12/console

jobs/integr8ly/browser-based-testsuite-pipeline.yaml
```diff
@@ -34,31 +34,47 @@
...
                         } 
                     }
                 }
 
                 stage('Publish test results') {
-                    dir('integreatly-qe/tests/ui-testsuite/reports') {
-                        archiveArtifacts '**/*.xml'                  
-                        junit allowEmptyResults:true, testResults: '**/*.xml'
-                    } 
+                    if (publishTestResults) {
```
matskiv: Do we still need this now? We have `allowEmptyResults:true` so I would guess that we don't need to handle empty reports folder.

pawelpaszki: you are right. Removed it

PR #45, opened 2019-02-12
Update Git checkout command to enable cloning from tags

tremes:
## Motivation
Current git command is not able to clone from tags.

### DIFF
--- a/jobs/integr8ly/installation-pipeline-qe-pony.yaml
+++ b/jobs/integr8ly/installation-pipeline-qe-pony.yaml
@@ -63,7 +63,13 @@
 
                 stage('Clone the installer') {
                     dir('installation') {
-                        git branch: BRANCH, url: REPOSITORY
+                        checkout([
+                            $class: 'GitSCM',
+                            branches: [[name: BRANCH]],,
+                            doGenerateSubmoduleConfigurations: false,
+                            extensions: [],
+                            userRemoteConfigs: [[url: REPOSITORY]]
+                        ])
                     } // dir
                 } // stage
 
--- a/jobs/integr8ly/installation-pipeline.yaml
+++ b/jobs/integr8ly/installation-pipeline.yaml
@@ -108,7 +108,13 @@
                 node(bastionLabel) {
                     stage('Clone the installer') {
                         dir('installation') {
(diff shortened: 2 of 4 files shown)


tremes:
@pawelpaszki @trepel  FYI

pawelpaszki:
@tremes did you try out any of those pipelines?

jobs/integr8ly/installation-pipeline-qe-pony.yaml
```diff
@@ -63,7 +63,13 @@
 
                 stage('Clone the installer') {
                     dir('installation') {
-                        git branch: BRANCH, url: REPOSITORY
+                        checkout([
+                            $class: 'GitSCM',
+                            branches: [[name: BRANCH]],,
+                            doGenerateSubmoduleConfigurations: false,
```
pawelpaszki: @tremes `doGenerateSubmoduleConfigurations` and `extensions` parameters seem to be repeating and providing no value (empty array and false). Would it be possible to remove those?

tremes:
@pawelpaszki Yes I needed to update our pipeline so that it's possible to install from tag.You can check https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/installation-pipeline/

jobs/integr8ly/installation-pipeline-qe-pony.yaml
tremes: Oh yes, it can be likely removed.

pawelpaszki: I have actually read about the `doGenerateSubmoduleConfigurations` and sometimes without setting it to false the build can fail, so we might have it here. Any thoughts @trepel ?

trepel: No thoughts @pawelpaszki . If so, let's have it there. I encountered this issue yesterday as well so I guess we should merge and update the jobs sooner rather than later.

pawelpaszki: Sure. lets merge it. @tremes do you want to update the job definitions?

PR #51, opened 2019-02-21
INTLY-564 first proposal for Nightly QE Pony pipeline

tremes:
@pawelpaszki @trepel Please review. This is first proposal for template for our nightly QE Pony pipeline (trigger pipeline needs to be created). It is created at https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/nightly-pipeline-qe-pony/ and the manually created trigger you can find in our "Nightly" tab. (`nightly-pipeline-qe-pony-trigger-new`).

I'll check the results tomorrow.

### DIFF
--- a/jobs/integr8ly/qe-poc-master-general.yaml
+++ b/jobs/integr8ly/qe-poc-master-general.yaml
@@ -1,0 +1,203 @@
+---
+
+- job:
+    name: qe-poc-master-general
+    project-type: pipeline
+    description: "Nightly pipeline for testing on POC (originally Pony) clusters."
+    sandbox: false
+    concurrent: true
+    properties:
+      - build-discarder:
+          num-to-keep: 20
+    parameters:
+        - string:
+            name: GH_CLIENT_ID
+            description: "GitHub OAuth Client ID, required for some of the walkthroughs. Can be left empty"
+        - string:
+            name: GH_CLIENT_SECRET
+            description: "GitHub OAuth Client Secret, required for some of the walkthroughs. Can be left empty"
+        - bool:
+            name: SELF_SIGNED_CERTS
+            default: false
+            description: "Indicates whether the cluster uses self signed certificates or not.
+This needs to be changed to 'false' for Integreatly Workshop since it uses valid certificates."
+        - string:
+            name: RECIPIENTS
+            default: integreatly-qe@redhat.com
+            description: "Whitespace- or comma-separated list of recipient addresses"
+        - string:
+            name: INSTALLATION_REPOSITORY
+            default: https://github.com/integr8ly/installation.git
+            description: "Repository of the Integreatly installer"
+        - string:
+            name: INSTALLATION_BRANCH
+            default: master
+            description: "Branch of the installer repository"
+        - string:
+            name: TEST_SUITES_REPOSITORY
+            default: https://gitlab.cee.redhat.com/integreatly-qe/integreatly-qe.git
+            description: "Repository of the Integreatly test suites"
+        - string:
+            name: TEST_SUITES_BRANCH
+            default: master
+            description: "Branch of a repository of Integreatly test suites"
+        - string:
+            name: CLUSTER_URL
+            description: "URL of cluster for installation."
+        - string:
+            name: MASTER_URLS
+            description: "Comma separated list of URLs for master nodes of target cluster to be used in Ansible inventory file"
+        - string:
+            name: ANSIBLE_USER
+            default: root
+            description: "User for Ansible to access the master node of target cluster"
+        - string:
+            name: OC_USER
+            default: trepel
+            description: "Default OpenShift admin user when Integr8ly is not installed."
+        - string:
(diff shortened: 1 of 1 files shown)


jobs/integr8ly/nightly-pipeline-qe-pony.yaml
```diff
@@ -0,0 +1,193 @@
...
+                            string(name: 'OC_PASSWORD', value: "${OC_PASSWORD}")]
+                                
+                            sleep time: 3, unit: 'MINUTES'
+                    } // stage
+                    
+                    stage('Execute "After Installation" Workarounds #2') {
+                        if (INSTALLATION_BRANCH == 'master') {
+                            // this patches the webapp to use master image (v2.2.0 is used otherwise)
+                            sh """
+                                oc login https://master.${YOURCITY}.openshiftworkshop.com -u admin@example.com -p Password1  --insecure-skip-tls-verify
```
pawelpaszki: I don't see **YOURCITY** in the parameter list

jobs/integr8ly/nightly-pipeline-qe-pony.yaml
```diff
@@ -0,0 +1,193 @@
...
+        - string:
+            name: OC_PASSWORD
+            description: "Default OpenShift admin password when Integr8ly is not installed."    
+    dsl: |
+        try {
+            timeout(180) { ansiColor('gnome-terminal') { timestamps {
+                node('cirhos_rhel7') {        
+                    stage('Verify input') {
+                        if (!MASTER_URLS) {
+                            throw new hudson.AbortException('MASTER_URLS parameter is required!')
```
pawelpaszki: Here only presence of one parameter is required, but there are more of them, which are required. I would either leave it out, or check if all required parameters are provided

jobs/integr8ly/nightly-pipeline-qe-pony.yaml
```diff
@@ -0,0 +1,193 @@
...
+                            
+                        if(buildStatus != 'SUCCESS') {
+                            currentBuild.result = 'UNSTABLE'
+                        }
+                            
+                        sleep time: 3, unit: 'MINUTES'
+                    } // stage
+
+                    stage('After First Login Tests') {
+                        buildStatus = build(job: 'after-first-login-tests', propagate: false, parameters: [
```
pawelpaszki: since recently there is a new parameter used for this pipeline: **EVALS_USERNAME**

jobs/integr8ly/nightly-pipeline-qe-pony.yaml
tremes: Good catch. THanks

jobs/integr8ly/nightly-pipeline-qe-pony.yaml
tremes: yeah this needs to be updated.

trepel:
@tremes Code looks good to me, I guess once the jobs pass on Jenkins, it is ok. One question though, I think we can have various integreatly-qe cluster related info as default values for the job parameters since I think it won't change too often. Even if the cluster is destroyed the new one will be re-created using the same URL and stuff. The situation we have two PoC clusters available is exceptional I guess

jobs/integr8ly/nightly-pipeline-qe-pony.yaml
tremes: I added some

tremes: I added some

tremes:
@trepel Yes you're likely right. I don't have any objections. We can hardcode more params in the future (if it is that what you mean). Thanks for review guys + @pawelpaszki

trepel:
@tremes I actually suggested to think about adding default values for CLUSTER_URL, MASTER_URLS. Since these will be pretty stable I think. Definitely not a blocker for approval.

trepel:
@tremes why is this called `nightly-pipeline-qe-pony`? In case of pds I named the job pds-general. We should unify our pipeline names so that the jobs serving the same purpose (but differ in cluster they target - PDS, PoC, OSD) have the same name schema. For instance: nightly-pds-heavy, nightly-qe-pony-heavy, nightly-osd-heavy etc.

The pds-general is not perfect name either. I would suggest pds-master-general (since it does not contain additional parameters required for relese testing and it contains workaround to patch images of webapp and walkthrouhgs to use master tag). So I would name this particular job as qe-pony-master-general.

Sorry for long comment, I hope it makes sense a bit.

tremes:
OK no problem but wouldn't be "qe-poc-master-general" better name then?:)

trepel:
yes, qe-poc-master-general is indeed better. Provided that we will make it master and general at some point. By master I mean that there will be patches webapp and walkthrough images to be 'master' and by general I mean that there will be choice parameter specifying what stages to execute (eg tests ony, or heavy, or (install + tests), etc.

PR #53, opened 2019-02-25
INTLY-850 attempt to improve our pipeline's reporting

tremes:
@trepel @pawelpaszki I tried to update our pipelines so that it receives proper buildStatus in case of e.g uninstallation failure. 
TBH I am not 100% how the buildStatus works (especially the propagate param - why the latest message says `propagate:false to ignore` when it wasn't set to false??) but with this it seems to be a little bit better.


tremes:
You can check latest emails in integreatly-qe. Nothing super but at least little bit better.

pawelpaszki:
it looks ok to me. WDYT @trepel

PR #57, opened 2019-02-25
Stop updating nightly PDS

trepel:
## Motivation & Why
Nightly PDS heavy job cannot be updated per night. The template contains placeholders instead of real value because the PDS is changing all the time - we need to provision new PDS cluser every three days

## How
By removing the job from configure_jenkins.sh

## Verification Steps
None, lets see if it is indeed skipped tomorrow.

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO


trepel:
@tremes Please have a look. Merging without review because I don't want this job to be regenerated at night.

tremes:
@trepel Oh yes. It's OK of course. I forgot we were discussing that.

PR #59, opened 2019-02-26
INTLY-852 fix bastionSlave removal

tremes:
There is an issue when we try to remove a non-existing bastion slave or a bastion slave which is executing some tasks. 

I experimented with this and it seems to work good. I executed two parallel jobs on my bastion slave. Originally I got a NPE because both of them were removing the slave every time (you can check https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/uninstallation-pipeline/722/console and the second job is  https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/tremes-testing-pipeline-INTLY-852/). After this fix both jobs worked as expected (not removing the bastion slave in finally block). 

Please check @trepel @pawelpaszki

### DIFF
--- a/jobs/integr8ly/installation-pipeline.yaml
+++ b/jobs/integr8ly/installation-pipeline.yaml
@@ -158,5 +158,10 @@
             // Clean up - remove jenkins slave that was created
         
             Node bastionSlave = Jenkins.instance.getNode(bastionLabel);
-            Jenkins.instance.removeNode(bastionSlave);
+            if(bastionSlave != null) {
+                Boolean isIdle = bastionSlave.toComputer().isIdle();
+                if(isIdle) {
+                    Jenkins.instance.removeNode(bastionSlave);
+                }
+            }
         }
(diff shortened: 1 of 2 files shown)


jobs/integr8ly/installation-pipeline.yaml
```diff
@@ -158,5 +158,8 @@
             // Clean up - remove jenkins slave that was created
         
             Node bastionSlave = Jenkins.instance.getNode(bastionLabel);
-            Jenkins.instance.removeNode(bastionSlave);
+            Boolean isIdle = bastionSlave.toComputer().isIdle();
```
trepel: I don't think that this solves the problem. If `bastionSlave` is null, then you get NPE on this line (calling `toComputer()` method on null object). The correct order should be:
1) null check
2) if not null see whether the slave is idle
2) remove the slave if it is not null and idle

tremes: ahh yes good catch. Thanks

trepel:
@tremes LGTM, approved. My personal taste would be to do just:
`if (bastionSlave != null && bastionSlave.toComputer().isIdle()) { ... }` but this is perfectly ok. Feel free to merge.

PR #61, opened 2019-02-27
Fix release creation for 1.2.x stream

mikenairn:
* Try and checkout a release branch name (v1.2) before falling back to master for each component.


mikenairn:
> I am not that familiar with this but it looks good to me. I would print some message in the catch block though. As is, it can silently use `releaseBaseBranch` while you will be expecting it using `releaseBranchName`

checkoutGitRepo will complain plenty when it tries to checkout a branch that doesn't exist, think it will be fine.

PR #70, opened 2019-03-06
add stage for fetching component product version

JameelB:
## Motivation
Automate the update of a component's product version (if different from the github release tag) when a new GA is released.
  
## What
Add a stage for fetching the component product version for the webapp.
- Separate the `manifestVar` into two variables for the product version and the release tag
- Add parameters for specifying where the product version is declared `productVersionLocation` and how it can be identified `productVersionIdentifier`
- The webapp doesn't have the correct information in the `WebappVersion` variable in the current [release](https://github.com/integr8ly/tutorial-web-app-operator/blob/v0.0.4/pkg/handlers/webhandler.go#L23). Instead, the product version can be retrieved from the [template](https://github.com/integr8ly/tutorial-web-app-operator/blob/v0.0.5/deploy/template/tutorial-web-app.yml#L71)

### DIFF
--- a/jobs/release-monitoring/discovery/github/Jenkinsfile
+++ b/jobs/release-monitoring/discovery/github/Jenkinsfile
@@ -209,11 +209,14 @@ def installationGitUrl = params.installationGitUrl ?: 'git@github.com:integr8ly/
 def installationGitRef = params.installationGitRef ?: 'master'
 def githubToken = params.githubToken ?: 'jenkins-github-api-token'
 def githubCredentialsID = params.credentialId ?: 'jenkinsgithub'
-def manifestVar = params.manifestVar
+def releaseTagVar = params.manifestVar ?: params.releaseTagVar
+def productVersionVar = params.productVersionVar
 def projectOrg = params.projectOrg
 def projectRepo = params.projectRepo
 def productName = params.productName
 def releaseFetchMethod = params.releaseFetchMethod
+def productVersionLocation = params.productVersionLocation
+def productVersionIdentifier = params.productVersionIdentifier
 def nextBranch = params.installationProductBranch ?: "${productName}-next"
 def installationManifestFile = './evals/inventories/group_vars/all/manifest.yaml'
 
@@ -227,7 +230,11 @@ node {
         dir('installation') {
             checkoutGitRepo(installationGitUrl, installationGitRef, githubCredentialsID)
             releaseConfig = readYaml file: installationManifestFile
-            componentRelease = releaseConfig[manifestVar]
+            if (releaseTagVar != "") {
+              componentRelease = releaseConfig[releaseTagVar]
+            } else {
+              componentRelease = releaseConfig[productVersionVar]
+            }
         }
     }
 
@@ -240,6 +247,47 @@ node {
         isGARelease = hasNewGARelease(componentRelease, latestRelease, productName)
         println "[INFO] latestRelease:${latestRelease}, componentRelease:${componentRelease}, isGARelease:${isGARelease}"
         currentBuild.description = "latest: ${latestRelease}\ncurrent: ${componentRelease}"
+    }
+
+    stage('Fetch Component Product Version') {
+        def productVersion = ""
+
+        // Only set the product version as the latest release if the release tag var is not available
+        if (!releaseTagVar) {
+          productVersion = latestRelease
+        }
+        // Ensures that old jobs will still work with old parameter
+        // Can be removed after all service discovery jobs have been updated
+        else if (params.manifestVar) {
+          productVersion = latestRelease
+        } else {
+          if (!productVersionLocation || !productVersionIdentifier) {
+            error "[ERROR] The product version location and/or identifier was not defined"
+          }
+
+          def templateUrl = "https://raw.githubusercontent.com/${projectOrg}/${projectRepo}/${latestRelease}/${productVersionLocation}"
+
+          if (projectOrg == "integr8ly") {
+            productVersion = sh(returnStdout: true, script: "curl -i '${templateUrl}' | grep '${productVersionIdentifier}' | head -n 1 | awk '{print \$3}'")
+            productVersion = productVersion.replaceAll("[\"\']v*", "")
+          }
+
+          // Current tags/releases does not have the right product version in the webapp handler.
+          // This can be removed once this property has been updated with the correct information in the next release
+          // https://github.com/integr8ly/tutorial-web-app-operator/blob/v0.0.4/pkg/handlers/webhandler.go#L23
+          if (productName == "webapp") {
+            productVersion = sh(returnStdout: true, script: "curl -i '${templateUrl}' | grep '${productVersionIdentifier}' | awk -F ':' '{print \$3}'")
+            productVersion = productVersion.replaceAll("v", "")
+          }
+
+          productVersion = productVersion.trim()
+
+          if (productVersion == "") {
+            error "[ERROR] Product version for ${productName} was not found. Product Version: ${productVersion}"
+          }
+        }
+
+        println "[INFO] product version: ${productVersion}"
     }
 
     dir('installation') {
@@ -260,8 +308,12 @@ node {
             stage('Product Version Update') {
                 when(isGARelease) {
                     gitCommitWhenChanges("Updated ${productName} product version to ${latestRelease}") { msgs ->
-                        manifestFileTxt = readFile(installationManifestFile)
-                        manifestFileTxt = manifestFileTxt.replaceFirst(/${manifestVar}: '.*'/, "${manifestVar}: '${latestRelease}'")
+                        manifestFileTxt = readFile(installationManifestFile)                        
+                        // Updates the release tag version
+                        manifestFileTxt = manifestFileTxt.replaceFirst(/${releaseTagVar}: '.*'/, "${releaseTagVar}: '${latestRelease}'")
+                        
+                        // Updates the product version
+                        manifestFileTxt = manifestFileTxt.replaceFirst(/${productVersionVar}: '.*'/, "${productVersionVar}: '${productVersion}'")
                         writeFile file: installationManifestFile, text: manifestFileTxt
                     }
                 }
--- a/jobs/release-monitoring/discovery/webapp.yaml
+++ b/jobs/release-monitoring/discovery/webapp.yaml
@@ -8,9 +8,14 @@
       - timed: '@hourly'
     parameters:
       - string:
-          name: 'manifestVar'
+          name: 'releaseTagVar'
           default: 'webapp_operator_release_tag'
-          description: '[REQUIRED] The manifest variable to be used as the current component version'
+          description: '[OPTIONAL] The manifest variable to be used as the current component release tag'
+          read-only: true
+      - string:
+          name: 'productVersionVar'
+          default: 'webapp_version'
+          description: '[REQUIRED] The manifest variable to be used as the current component product version'
           read-only: true
       - string:
           name: 'projectOrg'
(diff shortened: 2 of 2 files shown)


jobs/release-monitoring/discovery/github/Jenkinsfile
```diff
@@ -209,13 +209,17 @@ def installationGitUrl = params.installationGitUrl ?: 'git@github.com:integr8ly/
 def installationGitRef = params.installationGitRef ?: 'master'
 def githubToken = params.githubToken ?: 'jenkins-github-api-token'
 def githubCredentialsID = params.credentialId ?: 'jenkinsgithub'
-def manifestVar = params.manifestVar
+def releaseTagVar = params.releaseTagVar
+def productVersionVar = params.productVersionVar
```
mikenairn: Any way we can make this backward compatible so we can avoid breaking all the other jobs while we update them? Maybe we can set releaseTagVar to manifestVar if the param exists?

```
def releaseTagVar = params.manifestVar ?: params.releaseTagVar
```

jobs/release-monitoring/discovery/github/Jenkinsfile
```diff
@@ -240,6 +248,39 @@ node {
...
+          }
+
+          // Current tags/releases does not have the right product version in the webapp handler.
+          // This can be removed once this property has been updated with the correct information in the next release
+          // https://github.com/integr8ly/tutorial-web-app-operator/blob/v0.0.4/pkg/handlers/webhandler.go#L23
+          if (productName == "webapp") {
+            productVersion = sh(returnStdout: true, script: "curl -i '${templateUrl}' | grep '${productVersionIdentifier}' | awk -F ':' '{print \$3}'")
+          }
+
+          productVersion = productVersion.trim()
```
mikenairn: Product version for webapp is returning v<major>.<minor>.<patch>, where this variable previously only had a semver version string. Maybe we should remove any additional chars like the v.

jobs/release-monitoring/discovery/github/Jenkinsfile
```diff
@@ -227,7 +231,11 @@ node {
         dir('installation') {
             checkoutGitRepo(installationGitUrl, installationGitRef, githubCredentialsID)
             releaseConfig = readYaml file: installationManifestFile
-            componentRelease = releaseConfig[manifestVar]
+            if (releaseTagVar != "") {
+              componentRelease = releaseConfig[releaseTagVar]
+            } else {
+              componentRelease = releaseConfig[productVersionVar]
+            }
```
mikenairn: Bit unsure what this is for? Why do we use productVersionVar if releaseTagVar is empty? In what scenario would releaseTagVar be blank and why would we want to use productVersionVar instead in this case?

jobs/release-monitoring/discovery/github/Jenkinsfile
```diff
@@ -240,6 +248,39 @@ node {
         isGARelease = hasNewGARelease(componentRelease, latestRelease, productName)
         println "[INFO] latestRelease:${latestRelease}, componentRelease:${componentRelease}, isGARelease:${isGARelease}"
         currentBuild.description = "latest: ${latestRelease}\ncurrent: ${componentRelease}"
+
+        if (releaseTagVar == "") {
+          productVersion = latestRelease
+        }
```
mikenairn: Why set this to latestRelease? I see the productVersion and the currentRelease/latestRelease as two separate things, at least for the most part they are different anyway.

jobs/release-monitoring/discovery/github/Jenkinsfile
```diff
@@ -240,6 +248,39 @@ node {
...
+
+        println "[INFO] product version: ${productVersion}"
+    }
+
+    stage('Fetch Component Product Version') {
+        when(productVersion == "") {
```
mikenairn: Why do this if productVersion is blank? Maybe if we skipped this if the productVersionVar param was blank instead?

jobs/release-monitoring/discovery/github/Jenkinsfile
```diff
@@ -240,6 +248,39 @@ node {
...
+          println "[INFO] Retrieving component product version"
+          def templateUrl = "https://raw.githubusercontent.com/${projectOrg}/${projectRepo}/${latestRelease}/${productVersionLocation}"
+
+          if (projectOrg == "integr8ly") {
+            productVersion = sh(returnStdout: true, script: "curl -i '${templateUrl}' | grep '${productVersionIdentifier}' | head -n 1 | awk '{print \$3}'")
+            productVersion = productVersion.replaceAll("[\"\']", "")
+          }
```
mikenairn: Is this for other integeratly components other than webapp?

jobs/release-monitoring/discovery/github/Jenkinsfile
```diff
@@ -261,7 +302,12 @@ node {
                 when(isGARelease) {
                     gitCommitWhenChanges("Updated ${productName} product version to ${latestRelease}") { msgs ->
                         manifestFileTxt = readFile(installationManifestFile)
-                        manifestFileTxt = manifestFileTxt.replaceFirst(/${manifestVar}: '.*'/, "${manifestVar}: '${latestRelease}'")
+                        if (releaseTagVar != "") {
+                          // Updates the release tag version
+                          manifestFileTxt = manifestFileTxt.replaceFirst(/${releaseTagVar}: '.*'/, "${releaseTagVar}: '${latestRelease}'")
+                        }
+                        // Updates the product version
+                        manifestFileTxt = manifestFileTxt.replaceFirst(/${productVersionVar}: '.*'/, "${productVersionVar}: '${productVersion}'")
```
mikenairn: Since we are guarding against releaseTagVar being empty, although i'm not sure why that would ever be the case, should we add the same check around this update for productVersionVar?

jobs/release-monitoring/discovery/github/Jenkinsfile
JameelB: The release tags are only ever used for operators. Operator release tag will be different from the product version they install. In cases like `che` or `3scale` the 'release tag' they use in Github is their actual product version.

jobs/release-monitoring/discovery/github/Jenkinsfile
JameelB: yes, those should be for the other integreatly components. This could be used for the webapp in the future but at the moment, the value for the product version within that file in our current release is wrong.

jobs/release-monitoring/discovery/github/Jenkinsfile
JameelB: In some cases, the product version and the release tag are the same ([3Scale](https://github.com/integr8ly/installation/blob/master/evals/inventories/group_vars/all/manifest.yaml#L9), [che](https://github.com/integr8ly/installation/blob/master/evals/inventories/group_vars/all/manifest.yaml#L17)). Would it be best to split these in our manifest file? For each product, there should always be a release version and a product version, even if they have the same value?

mikenairn: OK, i understand where you're coming from now. I think what i would like is that we basically use releaseTagVar exactly like what manifestVar had been used for. It is a better representation of what we are passing in i think (the git tag to use). In the case where the releaseTagVar is the same as the product version (che, 3scale etc..), we just don't set productVersionVar, and in the 'Fetch Component Product Version' stage, if that is blank, we set it to the value of the release that was discovered.

This will mean the 2 stages are doing a job each:

'Fetch Latest GitHub Release' - Fetches the latest release based on releaseTagVar and sets latestRelease
'Fetch Component Product Version' - Fetches the latest product version based on productVersionVar and sets productVersion, but if productVersionVar is blank productVersion = latestVersion

jobs/release-monitoring/discovery/github/Jenkinsfile
```diff
@@ -209,13 +209,17 @@ def installationGitUrl = params.installationGitUrl ?: 'git@github.com:integr8ly/
...
 def githubToken = params.githubToken ?: 'jenkins-github-api-token'
 def githubCredentialsID = params.credentialId ?: 'jenkinsgithub'
-def manifestVar = params.manifestVar
+def releaseTagVar = params.manifestVar ?: params.releaseTagVar
+def productVersionVar = params.productVersionVar
 def projectOrg = params.projectOrg
 def projectRepo = params.projectRepo
 def productName = params.productName
 def releaseFetchMethod = params.releaseFetchMethod
+def templateLocation = params.productTemplateLocation
```
mikenairn: Is this used?

jobs/release-monitoring/discovery/github/Jenkinsfile
```diff
@@ -209,13 +209,17 @@ def installationGitUrl = params.installationGitUrl ?: 'git@github.com:integr8ly/
...
+def productVersionIdentifier = params.productVersionIdentifier
 def nextBranch = params.installationProductBranch ?: "${productName}-next"
 def installationManifestFile = './evals/inventories/group_vars/all/manifest.yaml'
+def productVersion = ""
```
mikenairn: Does this need to be declared here?

jobs/release-monitoring/discovery/github/Jenkinsfile
mikenairn: Do we still need this? Could we not set releaseTagVar to equal productVersionVar in the case where productVersionVar is the only thing we have?

```
def releaseTagVar =  params.manifestVar ? params.manifestVar : params.releaseTagVar ? params.releaseTagVar : params.productVersionVar
```

jobs/release-monitoring/discovery/github/Jenkinsfile
mikenairn: Can we not just skip this entire stage if "productVersionVar" isn't set?

```
if (productVersionVar) {
   //Do whatever it is we need to do if we have a productVersionVar 
   ...
} else {
   productVersion = latestRelease
}
```

jobs/release-monitoring/discovery/github/Jenkinsfile
```diff
@@ -249,6 +257,45 @@ node {
...
     }
 
+    stage('Fetch Component Product Version') {
+        // Only set the product version as the latest release if the release tag var is not available
+        if (!releaseTagVar) {
+          productVersion = latestRelease
+        }
+        // Ensures that old jobs will still work with old parameter
+        // Can be removed after all service discovery jobs have been updated
+        else if (params.manifestVar) {
```
mikenairn: Can't see why we would need this if we are setting the value releaseTagVar to manifestVar in the case where that's still being passed in.

jobs/release-monitoring/discovery/github/Jenkinsfile
```diff
@@ -249,6 +257,45 @@ node {
...
+          productVersion = latestRelease
+        } else {
+          if (!productVersionLocation || !productVersionIdentifier) {
+            println "[ERROR] The product version location and/or identifier was not defined"
```
mikenairn: Should this break here or something? I'm assuming if both of these are missing then it can't really do anything?

jobs/release-monitoring/discovery/github/Jenkinsfile
mikenairn: You seemed to have removed the check entirely? Is this correct? Surely this will fail if productVersionVar is blank which it will be for everything other than webapp?

jobs/release-monitoring/discovery/github/Jenkinsfile
JameelB: That should've been called `productVersionLocation`

jobs/release-monitoring/discovery/github/Jenkinsfile
JameelB: This is needed in order for the old jobs to set the product version correctly. At the moment, the `releaseTagVar` will always be defined. Manifest var will be there for the old jobs and the new webapp job will set that to a value for the release tag. If we don't catch this here, old jobs will try to go into the `else`.

PR #46, opened 2019-02-13
Remove evals/ dir from paths

matskiv:
## Motivation
`evals/` folder is being removed from the installer: https://github.com/integr8ly/installation/pull/353

## What
This PR updates paths in pipeline scripts.

## Verification Steps
Test syntax with jjb.

## Progress
- [ ] Update jobs in Jenkins once the [installer PR](https://github.com/integr8ly/installation/pull/353) is merged.

PR #71, opened 2019-03-06
Ensure installation evals directory exists

mikenairn:
* Add symlink to evals directory if one doesn't already exist.

The evals directory is being removed and everything will be in the root directory of the installation repo as expected see https://github.com/integr8ly/installation/pull/429. We however have a number of jobs that look for files in various places under evals/, this PR adds a symlink for evals after cloning the installation repo in appropriate jobs to ensure they continue to work.

When we are happy no jobs are dealing with old 1.2.x releases, we can update all the jobs to just reference the correct paths from the root directory. 

Note: This should be fine as long as jobs ensure the workspace is cleaned between jobs, and no jobs change the installation reference after the initial clone.

### DIFF
--- a/jobs/integr8ly/clean-uninstallation-pipeline.yaml
+++ b/jobs/integr8ly/clean-uninstallation-pipeline.yaml
@@ -27,6 +27,9 @@
                                         userRemoteConfigs: [[url: REPOSITORY]], 
                                         branches: [[name: BRANCH]]
                                         ]
+                            if(!fileExists("evals")) {
+                                sh "ln -s . evals"
+                            }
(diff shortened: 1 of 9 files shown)


mikenairn:
Wasn't aware of https://github.com/integr8ly/ci-cd/pull/46. This would be preferable, but i don't think we can just remove evals until we are sure these jobs don't need to deal with 1.2.x versions.

matskiv:
@trepel Sure. I closed it.

trepel:
@mikenairn @matskiv Thanks, merging this one.

PR #81, opened 2019-03-14
add service discovery job for backup container

JameelB:
## Motivation
Automate discovery for the [backup container image](https://github.com/integr8ly/backup-container-image/releases) releases.

## What
- Add discovery job for the backup container image releases
- Add branch job for the backup container image release to create the `backup-container-next` branch in the installation repo.

Test Release PR
- ~~https://github.com/integr8ly/installation/pull/456/files~~
- https://github.com/integr8ly/installation/pull/460

### DIFF
--- a/jobs/release-monitoring/discovery/backup-container.yaml
+++ b/jobs/release-monitoring/discovery/backup-container.yaml
@@ -1,0 +1,38 @@
+---
+- job:
+    name: release-monitoring-discovery/backup-container
+    display-name: 'Backup Container'
+    project-type: pipeline
+    concurrent: false
+    triggers:
+      - timed: '@hourly'
+    parameters:
+      - string:
+          name: 'productVersionVar'
+          default: 'backup_version'
+          description: '[REQUIRED] The manifest variable to be used as the current component version'
+          read-only: true
+      - string:
+          name: 'projectOrg'
+          default: 'integr8ly'
+          description: '[REQUIRED] github project organization'
+          read-only: true
+      - string:
+          name: 'projectRepo'
+          default: 'backup-container-image'
+          description: '[REQUIRED] github project repostirory'
+          read-only: true
+      - string:
+          name: 'productName'
+          default: 'backup-container'
+          description: '[REQUIRED] Product to check, this affects the way the job verifies if a new version if available'
+          read-only: true
+    pipeline-scm:
+      script-path: jobs/release-monitoring/discovery/github/Jenkinsfile
+      scm:
+        - git:
+            branches:
+              - master
+            url: 'https://github.com/integr8ly/ci-cd.git'
+            skip-tag: true
+            wipe-workspace: false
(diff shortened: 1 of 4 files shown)


jobs/release-monitoring/discovery/backup-container.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - string:
+          name: 'releaseTagVar'
+          default: 'backup_version'
+          description: '[OPTIONAL] The manifest variable to be used as the current component release tag'
+          read-only: true
+      - string:
+          name: 'productVersionVar'
+          default: 'backup_image_tag'
+          description: '[REQUIRED] The manifest variable to be used as the current component product version'
+          read-only: true
```
mikenairn: Are these 2 things not the wrong way round?

jobs/release-monitoring/discovery/github/Jenkinsfile
```diff
@@ -213,7 +213,7 @@ def ensureEvalsDir() {
 def updateManifestVariable(manifestFileTxt, name, value) {
     if (name && value) {
         println("Updating manifest variable: name = ${name}, value = ${value}")
-        manifestFileTxt = manifestFileTxt.replaceFirst(/${name}: '.*'/, "${name}: '${value}'")
+        manifestFileTxt = manifestFileTxt.replaceFirst(/${name}: .*/, "${name}: '${value}'")
```
mikenairn: What difference did this make?

jobs/release-monitoring/discovery/backup-container.yaml
mikenairn: Thinking about it, is there even a need for a backup_version variable? Can't see why there would be since it's just one of our operator and doesn't really operate a product with a version.

jobs/release-monitoring/discovery/github/Jenkinsfile
JameelB: When it has the `'`, it doesn't catch the versions if the version isn't surrounded by it. Version of the [backup](https://github.com/integr8ly/installation/blob/backup-container-next/inventories/group_vars/all/manifest.yaml#L95) is an example of this. Removing the `'` should now update versions  surrounded with or without `'`.

jobs/release-monitoring/discovery/backup-container.yaml
JameelB: From looking at the manifest file in the installer, it seems like the `backup_version` is being used as the release tag, and the `image_tag` is being used as the actual product version, the tag used for the image. I can switch these two if it makes more sense that way, i don't think it would be a problem.

JameelB: @mikenairn That's true. The image tag is being referenced in the [rhsso](https://github.com/integr8ly/installation/blob/master/roles/rhsso/defaults/main.yml#L57) but i don't see why two vars are needed. I think we could just use the `backup_version`. wdyt? @odra @pb82

pb82: I'd agree, the tag name and the image tag should have the same value for a release.

JameelB: @mikenairn @pb82 Created a PR to remove the `backup_image_tag` and use the `backup_version` instead. https://github.com/integr8ly/installation/pull/459

PR #85, opened 2019-03-19
Simplify uninstall pipeline as well

tremes:
We don't need to provide MASTER_URLS and BASTION_URL for RHPDS cluster because the format of these values is always the same.


tremes:
@trepel Can you please check and possibly merge?

PR #86, opened 2019-03-19
Smoke test pipeline - update input params

tremes:


tremes:
Yes it would be correct. On the other hand I introduced reasonable defautl values:)

trepel:
Hmm, ok, you got me. The `jenkins-jobs --conf jenkins_jobs.ini test integr8ly/installation-smoke-tests.yaml` is ok, LGTM then. Feel free to merge.

PR #87, opened 2019-03-19
Add service discovery for codeready

JameelB:
## Motivation
Automate service discovery for codeready

## What
- Add discovery job for codeready GA releases
  - Releases for codeready should be fetched by [tags](https://github.com/redhat-developer/codeready-workspaces-deprecated/tags).
  - Tags should be filtered by `GA` to only get the GA releases.
- Add branch job for the codeready release to create the codeready-next branch in the installation repo.
- Added param `gaReleaseTagRef` which will be used to filter out the `GA` releases from the repo's tags.

**NOTE**: The image tag for the codeready images are not locked down by a version in each release. They're tagged as `latest` by default ([1.0.0.GA](https://github.com/redhat-developer/codeready-workspaces-deprecated/blob/1.0.0.GA/operator-installer/deploy.sh#L68), [1.0.1.GA](https://github.com/redhat-developer/codeready-workspaces-deprecated/blob/1.0.1.GA/operator-installer/deploy.sh#L6)).

Test PR: https://github.com/integr8ly/installation/pull/467/files

### DIFF
--- a/jobs/release-monitoring/discovery/github/Jenkinsfile
+++ b/jobs/release-monitoring/discovery/github/Jenkinsfile
@@ -71,20 +71,17 @@ def gitFetch(url, gitTokenId) {
 }
 
 //Gets the latest release by git tags
-def getLatestReleaseByTag(gitOrg, gitRepo, gitTokenId, productName) {
+def getLatestReleaseByTag(gitOrg, gitRepo, gitTokenId, productName, gaReleaseTagRef) {
     def latestRelease = ""
     def url = "https://api.github.com/repos/${gitOrg}/${gitRepo}/git/refs/tags"
 
     def (code, releases) = gitFetch(url, gitTokenId)
     if (code != 200) {
-        error "[ERROR] Failed to retrieve tags for:${productName}. The :${gitRepo} repository does not have any tags available"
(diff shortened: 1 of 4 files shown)


jobs/release-monitoring/discovery/github/Jenkinsfile
```diff
@@ -72,19 +72,21 @@ def gitFetch(url, gitTokenId) {
...
-        error "[ERROR] Failed to retrieve tags for:${productName}. The :${gitRepo} repository does not have any tags available"
+      error "[ERROR] Failed to retrieve tags for:${productName}. The :${gitRepo} repository does not have any tags available"
     }
 
-    if (productName == "fuse") {
-        // Retrieve only Fuse's GA releases
-        releases = releases.findAll { release -> release.ref.contains("-redhat-") }
+    if (!gaReleaseTagRef) {
+      gaReleaseTagRef = ""
     }
```
mikenairn: You would be better to move this to outside the method and declare the variable with everything else here https://github.com/integr8ly/ci-cd/blob/master/jobs/release-monitoring/discovery/github/Jenkinsfile#L238 . You can also just declare it like this to avoid lines 84-85:

```
​def gaReleaseTagRef = params.gaReleaseTagRef ?: ""
```

PR #94, opened 2019-03-20
Remove label score

JameelB:
## What
- Remove `getLabelScore` from the Jenkins discovery pipeline.
  - The label passed to `getLabelScore()` is removed [here](https://github.com/integr8ly/ci-cd/blob/master/jobs/release-monitoring/discovery/github/Jenkinsfile#L148-L152) therefore this will always return a score of 0.
- Fix typos in the description of the productName and productRepo parameters
- Remove `ensureEvalsDir()`.
  - The evals directory no longer exists on the [installation](https://github.com/integr8ly/installation/commit/8af2bb98a3fe55fb3a9fa1766f845c39d1389121) repository.

### DIFF
--- a/jobs/release-monitoring/discovery/github/Jenkinsfile
+++ b/jobs/release-monitoring/discovery/github/Jenkinsfile
@@ -111,31 +111,6 @@ def getLatestRelease(gitOrg, gitRepo, gitTokenId, releaseFetchMethod, gaReleaseT
     }
 }
 
-def getLabelScore(label) {
-    if (!label) {
-        return 0
-    }
-
-    def score = 0
-    def scores = [
-            'GA': 1,
-            'redhat': 1
-    ]
-    def parts = label.tokenize('-')
-
-    parts.each { part ->
-        if (scores.containsKey(part)) {
-            score += scores[part]
-        }
-
-        if (!scores.containsKey(part) && part.isInteger()) {
-            score += part.toInteger()
-        }
-    }
-
-    return score
-}
-
 //checks if current semver is lower than the supposed latest one
 def hasNewGARelease(currentVersion, newVersion, productName) {
     def current = ""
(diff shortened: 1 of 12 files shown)


jobs/release-monitoring/discovery/github/Jenkinsfile
```diff
@@ -246,7 +206,6 @@ node {
         cleanWs()
         dir('installation') {
             checkoutGitRepo(installationGitUrl, installationGitRef, githubCredentialsID)
-            ensureEvalsDir()
```
JameelB: Would there be any issues if we removed this? @mikenairn

mikenairn: It probably isn't actually required in here as long as we are never dealing with any branch other than master i.e v1.2. I can't see that ever happening, so should be OK to remove now.

jobs/release-monitoring/discovery/github/Jenkinsfile
```diff
@@ -114,31 +114,6 @@ def getLatestRelease(gitOrg, gitRepo, gitTokenId, releaseFetchMethod) {
...
-        }
-
-        if (!scores.containsKey(part) && part.isInteger()) {
-            score += part.toInteger()
-        }
-    }
-
-    return score
-}
-
```
mikenairn: Why is this not needed anymore? Not really sure what it was doing initially @odra added it. Think it was supposed to figure out if a tag indicated a new GA release or something.

JameelB: The way we're getting the [current/latest ](https://github.com/integr8ly/ci-cd/blob/master/jobs/release-monitoring/discovery/github/Jenkinsfile#L148-L152)version right now is by removing all text apart from the semver. This includes removing the label to indicate that it's a GA release. This function isn't doing anything for us at the moment because the [value](https://github.com/integr8ly/ci-cd/blob/master/jobs/release-monitoring/discovery/github/Jenkinsfile#L182) that's being passed to that function will always be null. 

We are [filtering](https://github.com/integr8ly/ci-cd/blob/master/jobs/release-monitoring/discovery/github/Jenkinsfile#L85) for the GA releases when we get releases by tags so I don't think we need that function at the moment.

PR #95, opened 2019-03-20
INTLY-1350 - Add openshift cluster create job

mikenairn:
Jira:

* https://issues.jboss.org/browse/INTLY-1350

### DIFF
--- a/jobs/openshift/cluster/create/Jenkinsfile
+++ b/jobs/openshift/cluster/create/Jenkinsfile
@@ -3,7 +3,7 @@
 def clusterOptions = [:]
 
 def verifyClusterOptions(clusterOptions) {
-    if(!clusterOptions.name || !clusterOptions.awsRegion) {
+    if (!clusterOptions.name || !clusterOptions.awsRegion) {
         def userInput = input message: 'Cluster Options', parameters: [
                 string(defaultValue: (clusterOptions.name ?: ''), description: 'Cluster name', name: 'clusterName'),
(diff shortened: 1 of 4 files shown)


jobs/openshift/cluster/create/Jenkinsfile
```diff
@@ -0,0 +1,28 @@
...
+        wrap([$class: 'AnsiColorBuildWrapper', colorMapName: "xterm"]) {
+            ansibleTower(
+                    towerServer: 'Dev Tower',
+                    jobTemplate: '[aws.eng] Cluster Create',
+                    templateType: 'workflow',
+                    importTowerLogs: true,
+                    removeColor: false,
+                    verbose: true,
+                    extraVars: """---
+                    oo_clusterid: ${clusterName}
```
JameelB: Does this generate an id if it's an empty string?

Maybe we should set a default value for `clusterName` in case the param isn't defined.

```
def clusterName = params.clusterName ?: ''
```

mikenairn: The default value is an empty string, so no need to default it like that. That said, it should probably check for an empty string and not continue if it is, but we should probably also be doing that in lots of jobs.

mikenairn: I added an input verification stage https://github.com/integr8ly/ci-cd/pull/95/commits/21d817e67e2ecf152ec139d0fd30c4742518a1c0, this is more like what we should be doing.

PR #98, opened 2019-03-21
added pipeline to execute enmasse address creation test

pawelpaszki:
## Motivation
https://issues.jboss.org/browse/INTLY-1389

## Verification Steps
1. Go to https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/Integr8ly/job/enmasse-address-creation-test/4/console and check the output of the pipeline
2. Inspect the yaml file
3. alterntively run the pipeline: https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/Integr8ly/job/enmasse-address-creation-test/build?delay=0sec with your own params (NOTE: it is required that the evals user you provide has already logged in into the webapp)

## Checklist:

## Progress

- [x] Finished task

### DIFF
--- a/jobs/integr8ly/enmasse-address-creation-test.yaml
+++ b/jobs/integr8ly/enmasse-address-creation-test.yaml
@@ -1,0 +1,73 @@
+---
+
+- job:
+    name: enmasse-address-creation-test
+    description: 'Executes enmasse address creation test'
+    project-type: pipeline
+    sandbox: true
+    parameters:
+      - string:
+          name: REPOSITORY
+          default: 'https://gitlab.cee.redhat.com/integreatly-qe/integreatly-qe.git'
+          description: 'QE repository containing the test source code.'
+      - string:
+          name: BRANCH
+          default: 'master'
+          description: 'Branch of the repository'
+      - string:
+          name: CLUSTER_URL
+          description: 'URL of cluster on which the test will be executed.'
+      - string:
+          name: ADMIN_USERNAME
+          default: 'admin@example.com'
+          description: 'Username to login to Integreatly cluster.'
+      - string:
+          name: ADMIN_PASSWORD
+          default: 'Password1'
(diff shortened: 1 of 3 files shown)


jobs/integr8ly/enmasse-address-creation-test.yaml
```diff
@@ -0,0 +1,73 @@
+---
+
+- job:
+    name: enmasse-address-creation-test
```
tremes: Maybe better to use AMQ Online? WDYT? I don't have a strong opinion here but this might be confusing for newcomers

pawelpaszki: done. It can be verified here: https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/Integr8ly/job/amq-online-address-creation-test/2/console ( I will probably need to change the test to wait 90 instead of 60 seconds for the address to be ready, because every now and then the test itself fails (not related to the pipeline))

tremes: ok. I was thinking about some global TIMEOUT variable in our testsuite (so that it can be increased globally) and we can e.g multiply this value in each test accordingly. I think I will sum it up in some JIRA

tremes: Thanks anyway

pawelpaszki: so are there any other changes required for this PR?

tremes: One more update - add the new pipeline to the script for pipelines recreation please.

pawelpaszki: Good catch! Done

PR #101, opened 2019-03-22
Add job for installing integreatly on OSD via ansible tower

JameelB:
## What
**Installation**
Add a Jenkins job for triggering the `Integreatly Install Workflow` template in Ansible Tower.
Parameters required for the survey of the `Integreatly Install Workflow` template:
- cluster_name
- eval_seed_users_count
- eval_self_signed_certs
- integreatly_inventory_source_aws_credentials
- integreatly_project_bootstrap_cluster_scm_branch
- integreatly_project_bootstrap_cluster_scm_url
- openshift_master_public_url
- openshift_password
- openshift_username


tremes:
We already have one simple pipeline for installation - https://github.com/integr8ly/ci-cd/blob/master/jobs/integr8ly/installation-pipeline.yaml Maybe we can unify it?

JameelB:
@tremes This is using ansible tower to install integreatly on OSD. The one you linked above installs integreatly on PDS. I think it would be difficult to unify both as they would need different parameters.

tremes:
@JameelB Ahh ok. Then it doesn't make sense. Thanks for clarification.

PR #104, opened 2019-03-26
ensure job fails on a failed rebase of product next branch

JameelB:
## What
The Jenkins job for the release discovery succeeds even if the `<product>-branch` fails to rebase from master. The job should fail when this happens to ensure that we are able to catch and fix merge conflicts.


JameelB:
Closing, will revisit this later.

PR #108, opened 2019-03-28
INTLY-1516 create a pipeline for users creation

tremes:
Successfull run https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/sso-user-create-tests/

### DIFF
--- a/jobs/integr8ly/sso-user-create.yaml
+++ b/jobs/integr8ly/sso-user-create.yaml
@@ -1,0 +1,75 @@
+---
+
+- job:
+    name: sso-user-create-tests
+    project-type: pipeline
+    sandbox: true
+    parameters:
+      - string:
+          name: REPOSITORY
+          default: 'https://gitlab.cee.redhat.com/integreatly-qe/integreatly-qe.git'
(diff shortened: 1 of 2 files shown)


tremes:
You can try to run it against OSD cluster.

jobs/integr8ly/sso-user-create.yaml
```diff
@@ -0,0 +1,67 @@
...
+      - string:
+          name: SSO_URL
+          description: 'SSO URL for the cluster on which the tests will be executed.'
+      - bool:
+          name: CLEAN_RESOURCES_USERS
+          default: true
+          description: 'Whether the users should be deleted after the successful execution or not.'
+      - string:
+          name: NUMBER_OF_USERS
+          default: "5"
```
pawelpaszki: one minor thing - I think we should make the default number consistent with the test code (e.g. 9). Other than that LGTM

tremes: OK so I would update the test source code then since 9 seems to me too many users.:)

PR #109, opened 2019-04-02
Update cluster create job to reflect changes in tower configuration.

mikenairn:
## What

* Update cluster create job to reflect changes in tower configuration.


mikenairn:
@pmccarthy Making you aware of these jobs more than anything.

PR #117, opened 2019-04-04
Add job for uninstalling integreatly on OSD via ansible tower

JameelB:
## What
Add a Jenkins job for uninstalling integreatly on OSD via ansible tower. This Job will call the workflow `Integreatly Uninstall Template` in the Ansible Tower dev instance with the following parameters: 
  - cluster_name
  - openshift_master_public_url
  - openshift_username
  - openshift_password
  - integreatly_project_uninstall_scm_url
  - integreatly_project_uninstall_scm_branch

### DIFF
--- a/jobs/openshift/cluster/integreatly/openshift-cluster-integreatly-uninstall.yaml
+++ b/jobs/openshift/cluster/integreatly/openshift-cluster-integreatly-uninstall.yaml
@@ -1,0 +1,43 @@
+---
+- job:
+    name: openshift-cluster-integreatly-uninstall
+    display-name: 'Openshift Cluster Integreatly Uninstall'
+    project-type: pipeline
+    parameters:
+      - string:
+          name: 'clusterName'
+          default: ''
+          description: '[REQUIRED] The name of the target cluster to uninstall Integreatly against'
+      - string:
+          name: 'openshiftMasterUrl'
+          default: ''
+          description: '[REQUIRED] The public URL of the target OpenShift cluster'
+      - string:
+          name: 'clusterAdminUsername'
+          default: 'integreatly'
+          description: '[REQUIRED] Username of the cluster admin account'
+      - string:
+          name: 'clusterAdminPassword'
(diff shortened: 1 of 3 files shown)


jobs/openshift/cluster/integreatly/openshift-cluster-integreatly-uninstall.yaml
```diff
@@ -0,0 +1,43 @@
...
+          name: 'openshiftMasterUrl'
+          default: ''
+          description: '[REQUIRED] The public URL of the target OpenShift cluster'
+      - string:
+          name: 'clusterAdminUsername'
+          default: 'integreatly'
+          description: '[REQUIRED] Username of the cluster admin account'
+      - string:
+          name: 'clusterAdminPassword'
+          default: 'Password1'
```
mikenairn: This will need changed as part of the password update. Can be done after merge.

jobs/openshift/cluster/integreatly/openshift-cluster-integreatly-uninstall.yaml
```diff
@@ -0,0 +1,43 @@
...
+          description: '[REQUIRED] Password of the cluster admin account'
+      - string:
+          name: 'uninstallerGitUrl'
```
mikenairn: I think this should just be installationGitUrl like the other jobs. It's not really because it was in the install job, more that it is the installation repo url.

jobs/openshift/cluster/integreatly/openshift-cluster-integreatly-uninstall.yaml
```diff
@@ -0,0 +1,43 @@
...
+          default: 'https://github.com/integr8ly/installation.git'
+          description: '[REQUIRED] Integreatly uninstaller Git URL'
+      - string:
+          name: 'uninstallerGitBranch'
```
mikenairn: Same here.

PR #125, opened 2019-04-09
changed cluster_url param in poc nightly trigger to yourcity

pawelpaszki:
## What
changed cluster_url param in poc nightly trigger to yourcity. 

The poc-testing-nightly-trigger job, which is triggering poc-testing-executor is running here: https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/Nightly/job/poc-testing-nightly-trigger/66/console

### DIFF
--- a/jobs/integr8ly/poc-testing-executor.yaml
+++ b/jobs/integr8ly/poc-testing-executor.yaml
@@ -42,8 +42,9 @@ This needs to be changed to 'false' for Integreatly Workshop since it uses valid
             default: master
             description: "Branch of a repository of Integreatly test suites"
         - string:
-            name: CLUSTER_URL
-            description: "URL of cluster for installation."
+            name: YOURCITY
+            default: integreatly-qe.opentry.me
+            description: "City or Customer (5 char min.) plus the generated hash, e.g. qebrno-5d10 [required]"
         - string:
             name: MASTER_URLS
             description: "Comma separated list of URLs for master nodes of target cluster to be used in Ansible inventory file"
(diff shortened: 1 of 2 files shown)


jobs/integr8ly/poc-testing-executor.yaml
```diff
@@ -42,8 +42,9 @@ This needs to be changed to 'false' for Integreatly Workshop since it uses valid
             default: master
             description: "Branch of a repository of Integreatly test suites"
         - string:
-            name: CLUSTER_URL
-            description: "URL of cluster for installation."
+            name: YOURCITY
```
trepel: I know we discussed this on chat but looking at the PR I changed my mind (sorry). The `YOURCITY` is well-defined thing in PDS context but I don't think it is a good name in POC context. What we actually need is the cluster URL without the protocol. I think the best name would be CLUSTER_HOST (or CLUSTER_HOSTNAME)

jobs/integr8ly/poc-testing-executor.yaml
```diff
@@ -42,8 +42,9 @@ This needs to be changed to 'false' for Integreatly Workshop since it uses valid
...
+            default: integreatly-qe.opentry.me
+            description: "City or Customer (5 char min.) plus the generated hash, e.g. qebrno-5d10 [required]"
```
trepel: The description here is not correct. Also, it should be changed if you agree with my previous comment. My proposal would be "Hostname of the cluster. Typically it is just a cluster URL without the protocol part [required]"

pawelpaszki:
@trepel changed the param name. The job is about to complete (with `YOURCITY` param though). Should we re-run it?

PR #127, opened 2019-04-10
add pipeline for testing integreatly on poc clusters

JameelB:
## What
Adds a pipeline for testing Integreatly on an OSD cluster. This job should include provisioning and deprovisioning a new cluster as well as installing, testing and uninstalling Integreatly.

- [x] Add an integreatly test job within `openshift/cluster/integreatly/`
- [x] Ensure job accepts the installation repo's git url and branch as parameters
- [x] Create a Jenkinsfile for this pipeline with the following stages:
  - **Create Cluster**: Should call the `openshift-cluster-create` pipeline
  - **Install Integreatly**: Should call the `openshift-cluster-integreatly-install` pipeline
  - **Test Integreatly**: An empty stage for now, this will be done as part of a separate [ticket](https://issues.jboss.org/browse/INTLY-1700)
  - **Uninstall Integreatly**: Should call the `openshift-cluster-integreatly-uninstall` pipeline
  - **Deprovision Cluster**: Should call the `openshift-cluster-deprovision` pipeline

### DIFF
--- a/jobs/openshift/cluster/integreatly/openshift-cluster-integreatly-test.yaml
+++ b/jobs/openshift/cluster/integreatly/openshift-cluster-integreatly-test.yaml
@@ -1,0 +1,27 @@
+---
+- job:
+    name: openshift-cluster-integreatly-test
+    display-name: 'Openshift Cluster Integreatly Test'
+    project-type: pipeline
+    parameters:
+      - string:
+          name: 'installerGitUrl'
+          default: 'https://github.com/integr8ly/installation.git'
+          description: '[REQUIRED] Integreatly installer Git URL'
+      - string:
+          name: 'installerGitBranch'
+          default: ''
+          description: '[REQUIRED] The name of the git branch to be used for installing Integreatly'
+      - bool:
+          name: 'dryRun'
+          default: false
+          description: '[OPTIONAL][Test] Dry run only, only prints what it would do!'
+    pipeline-scm:
+      script-path: jobs/openshift/cluster/integreatly/test/Jenkinsfile
+      scm:
+        - git:
+            branches:
+              - 'master'
+            url: 'https://github.com/integr8ly/ci-cd.git'
+            skip-tag: true
+            wipe-workspace: false
--- a/jobs/openshift/cluster/integreatly/openshift-cluster-integreatly-test.yaml
+++ b/jobs/openshift/cluster/integreatly/openshift-cluster-integreatly-test.yaml
@@ -10,7 +10,7 @@
           description: '[REQUIRED] Integreatly installer Git URL'
       - string:
           name: 'installerGitBranch'
-          default: ''
+          default: 'master'
           description: '[REQUIRED] The name of the git branch to be used for installing Integreatly'
       - bool:
           name: 'dryRun'
--- a/jobs/openshift/cluster/integreatly/test/Jenkinsfile
+++ b/jobs/openshift/cluster/integreatly/test/Jenkinsfile
@@ -1,5 +1,3 @@
-#!groovy
-
 def testOptions = [:]
 
 def verifyParameters(testOptions) {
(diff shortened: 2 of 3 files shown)


jobs/openshift/cluster/integreatly/openshift-cluster-integreatly-test.yaml
```diff
@@ -0,0 +1,27 @@
...
+    display-name: 'Openshift Cluster Integreatly Test'
+    project-type: pipeline
+    parameters:
+      - string:
+          name: 'installerGitUrl'
+          default: 'https://github.com/integr8ly/installation.git'
+          description: '[REQUIRED] Integreatly installer Git URL'
+      - string:
+          name: 'installerGitBranch'
+          default: ''
```
odra: shouldn't we use "master" as default?

JameelB: I was considering master but I think this job will mostly be run for the suites branch: integreatly-next and integration-next, so I wasn't sure what default value to put in here.

mikenairn: I don;t mind either way tbh, set a default of master if we think it's useful.

jobs/openshift/cluster/integreatly/test/Jenkinsfile
```diff
@@ -0,0 +1,109 @@
...
+  verifyParameters(testOptions)
+}
+
+node('cirhos_rhel7') {
+  cleanWs()
+  stage('Create Cluster') {
+    println "[INFO] Creating ${testOptions.clusterName} cluster"
+    def jobName = 'openshift-cluster-create'
+    def instanceType = 'm4.xlarge'
+    def instanceGroupSize = '3'
```
mikenairn: Probably wan't to adjust these once we have the ability to reduce the size of the clusters. Issue on the backlog to look into why that's not working properly at the moment.

mikenairn: Probably won;t all be the same either, master should be 1 and infra and compute 2 or more. Should we consider just setting the default values in the create job instead of passing them in here?

JameelB: The current default value of these is set to 3 in our repo but set to the values you mentioned above on Jenkins. Should these be changed to: master(1), infra/compute(2)? I can include that change in this PR. 

If the configuration above would be enough, it would be great to use the default values.

jobs/openshift/cluster/integreatly/test/Jenkinsfile
```diff
@@ -0,0 +1,109 @@
...
+      [$class: 'StringParameterValue', name: 'installerGitBranch', value: testOptions.gitBranch],
+      [$class: 'StringParameterValue', name: 'userCount', value: userCount],
+      [$class: 'BooleanParameterValue', name: 'selfSignedCerts', value: selfSignedCerts],
+      [$class: 'BooleanParameterValue', name: 'dryRun', value: params.dryRun]
+    ]
+    buildJob(jobName, jobParams)
+  }
+
+  stage('Test Integreatly') {
+    println "[INFO] Testing Integreatly"
```
mikenairn: Jira on the backlog to cover implementing this stage https://issues.jboss.org/browse/INTLY-1700. Will probably need sync up with @tremes et al on what integration testing is in place already.

jobs/openshift/cluster/integreatly/test/Jenkinsfile
```diff
@@ -0,0 +1,109 @@
...
+    def jobName = 'openshift-cluster-deprovision'
+    def jobParams = [
+      [$class: 'StringParameterValue', name: 'clusterName', value: testOptions.clusterName],
+      [$class: 'StringParameterValue', name: 'awsRegion', value: testOptions.awsRegion],
+      [$class: 'StringParameterValue', name: 'awsAccountName', value: testOptions.awsAccountName],
+      [$class: 'StringParameterValue', name: 'clusterDomainName', value: testOptions.domainName],
+      [$class: 'BooleanParameterValue', name: 'dryRun', value: params.dryRun]
+    ]
+    buildJob(jobName, jobParams)
+  }
```
mikenairn: @JameelB We need to ensure this stage always gets executed at the end regardless of the status of any of the earlier stages. I think unfortunately the only way to do that in a scripted pipeline is wrapping everything a try finally block. If you felt like you could also try a declarative pipeline instead which provides you with a post/always option.

jobs/openshift/cluster/integreatly/test/Jenkinsfile
tremes: The problem with testing is that we heavily rely on admin permissions. I've heard about some dedicated-admin account, but it still doesn't sound much promising.

jobs/openshift/cluster/integreatly/test/Jenkinsfile
mikenairn: I'm thinking we should make the default values in the create job be the minimum that we will require the majority of the time. So probably master(1), infra/compute(2) as you say, so if you just remove all those params from here, when we do eventually update that job (Once we figure out https://issues.jboss.org/browse/INTLY-1675), this job will just do the correct thing.

jobs/openshift/cluster/integreatly/test/Jenkinsfile
mikenairn: @tremes OK thanks. We will need to run integration testing since it's one of the requirements. If you have any comments/suggestions/thoughts about any of that, please add it into that ticket.

tremes: I was obviously confused here. I thought we were talking about the OSD testing. I think we should be OK with the clusters provisioned by ansible tower tooling since there's admin account available.

jobs/openshift/cluster/create/openshift-cluster-create.yaml
```diff
@@ -35,15 +35,15 @@
...
           description: '[REQUIRED] The amount of AWS master nodes to be created'
       - string:
           name: 'compute_group_size'
-          default: '3'
+          default: '2'
           description: '[REQUIRED] The amount of AWS compute nodes to be created'
       - string:
           name: 'infra_group_size'
-          default: '3'
+          default: '2'
```
mikenairn: I'd leave these values as they were for the time being, we have a small issue to resolve before we can change these https://issues.jboss.org/browse/INTLY-1675. We also probably need the compute size to be 3.

mikenairn:
@JameelB Looks good to me, but i guess we should verify it all works when @pmccarthy get's done updating the dev tower. I'd be amazed if it all worked first time :crossed_fingers:

jobs/openshift/cluster/integreatly/test/Jenkinsfile
```diff
@@ -0,0 +1,124 @@
...
+    stage('Uninstall Integreatly') {
+      steps {
+        echo "[INFO] Uninstalling Integreatly from ${testOptions.clusterName}"
+        script {
+          jobName = 'openshift-cluster-integreatly-uninstall'
+          jobParams = [
+            [$class: 'StringParameterValue', name: 'clusterName', value: testOptions.clusterName],
+            [$class: 'StringParameterValue', name: 'openshiftMasterUrl', value: testOptions.openshiftMasterUrl],
+            [$class: 'StringParameterValue', name: 'uninstallerGitUrl', value: testOptions.gitUrl],
+            [$class: 'StringParameterValue', name: 'uninstallerGitBranch', value: testOptions.gitBranch],
```
mikenairn: If we rename these here https://github.com/integr8ly/ci-cd/pull/117#discussion_r274423390, they will need changed here also.

jobs/openshift/cluster/integreatly/openshift-cluster-integreatly-test.yaml
```diff
@@ -0,0 +1,27 @@
+---
+- job:
+    name: openshift-cluster-integreatly-test
+    display-name: 'Openshift Cluster Integreatly Test'
+    project-type: pipeline
+    parameters:
+      - string:
+          name: 'installerGitUrl'
```
mikenairn: Just noticed this is called installer instead of installation. I think in almost all other job defs we have called this installationGitUrl as it represents the installation repo (https://github.com/integr8ly/installation/). Probably best to keep it consistent.

jobs/openshift/cluster/integreatly/test/Jenkinsfile
mikenairn: Will address the param naming in a follow up PR.

jobs/openshift/cluster/integreatly/test/Jenkinsfile
```diff
@@ -0,0 +1,124 @@
...
+    node {
+      label 'cirhos_rhel7'
+    }
+  }
+
+  parameters {
+    string(defaultValue: 'https://github.com/integr8ly/installation.git', description: '[REQUIRED] Integreatly installer Git URL', name: 'installerGitUrl')
+    string(defaultValue: 'master', description: '[REQUIRED] The name of the git branch to be used for installing Integreatly', name: 'installerGitBranch')
+    booleanParam(defaultValue: false, description: '[OPTIONAL][Test] Dry run only, only prints what it would do!', name: 'dryRun')
+  }
```
mikenairn: @JameelB Do you know if these are required here if we already have them in the job definition above?

mikenairn: Can look into removing this duplication after.

PR #128, opened 2019-04-10
Minor updates of the testing pipelines caused by my work on INTLY-154…

tremes:


jobs/integr8ly/pds-testing-executor.yaml
```diff
@@ -168,6 +168,7 @@ This needs to be changed to 'false' for Integreatly Workshop since it uses valid
 
             if (testPipeline == 'after-first-login-tests' || testPipeline == 'browser-based-testsuite-pipeline') {
                 parameters.add(string(name: 'EVALS_USERNAME', value: 'evals23@example.com'))
+                parameters.add(string(name: 'EVALS_PASSWORD', value: 'Password1'))
```
pawelpaszki: I think you should additionally remove the EVALS_PASSWORD from the next 'if statement'

jobs/integr8ly/poc-testing-executor.yaml
```diff
@@ -200,6 +200,7 @@ This needs to be changed to 'false' for Integreatly Workshop since it uses valid
 
             if (testPipeline == 'after-first-login-tests' || testPipeline == 'browser-based-testsuite-pipeline') {
                 parameters.add(string(name: 'EVALS_USERNAME', value: 'evals11@example.com'))
+                parameters.add(string(name: 'EVALS_PASSWORD', value: 'Password1'))
```
pawelpaszki: I think you should additionally remove the EVALS_PASSWORD from the next 'if statement'

jobs/integr8ly/pds-testing-executor.yaml
tremes: OK good point I think I will rather remove the or condition (I think it will be more clear)

PR #133, opened 2019-04-11
use jenkins credentials for the integreatly cluster username and pw

JameelB:
## What
Use the credentials stored in Jenkins called `tower-openshift-cluster-credential` which contains the admin username and password of the cluster provisioned by Ansible Tower.

### DIFF
--- a/jobs/openshift/cluster/integreatly/openshift-cluster-integreatly-install.yaml
+++ b/jobs/openshift/cluster/integreatly/openshift-cluster-integreatly-install.yaml
@@ -14,12 +14,12 @@
           description: '[REQUIRED] The public URL of the target OpenShift cluster'
       - string:
           name: 'clusterAdminUsername'
-          default: 'integreatly'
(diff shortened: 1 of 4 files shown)


jobs/openshift/cluster/integreatly/openshift-cluster-integreatly-install.yaml
```diff
@@ -12,14 +12,6 @@
           name: 'openshiftMasterUrl'
           default: ''
           description: '[REQUIRED] The public URL of the target OpenShift cluster'
-      - string:
```
JameelB: @mikenairn @odra Do we really need this as a parameter here? I've removed it because we don't really have a choice on what credentials are created on the cluster during provision.

mikenairn: Probably not as long as the clusters we are testing against always use the same user/password. If we ever had to test on a cluster that wasn't created by tower, this could be different.

PR #135, opened 2019-04-15
Intly 1675 group size

odra:
## What

* new vars in create cluster job definition file to use desired size vars for nodes
* is sets  the min size var to "1" for compute and infra nodes if the desired size number is "1"

### DIFF
--- a/jobs/openshift/cluster/create/Jenkinsfile
+++ b/jobs/openshift/cluster/create/Jenkinsfile
@@ -61,12 +61,12 @@ node('cirhos_rhel7') {
                     openshift_aws_infra_group_desired_size: ${clusterOptions.nodes.infra.group_size}"""
             if ((clusterOptions.nodes.compute.group_size as Integer) == 1) {
                 extraVars = """${extraVars}
-                openshift_aws_compute_group_min_size=1"""
+                    openshift_aws_compute_group_min_size:1"""
             }
 
             if ((clusterOptions.nodes.infra.group_size as Integer) == 1) {
                 extraVars = """${extraVars}
-                openshift_aws_infra_group_desired_size=1"""
+                    openshift_aws_infra_group_desired_size:1"""
             }
 
             println extraVars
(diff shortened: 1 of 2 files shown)


jobs/openshift/cluster/create/Jenkinsfile
```diff
@@ -48,6 +48,30 @@ node('cirhos_rhel7') {
...
+                "openshift_aws_infra_group_desired_size: ${clusterOptions.nodes.infra.group_size}"
+            ]
+            
+            if (clusterOptions.nodes.compute.group_size == '1') {
+                extraVars << 'openshift_aws_compute_group_min_size: 1'
+            }
+
+            if (clusterOptions.nodes.infra.group_size == '1') {
+                extraVars << 'openshift_aws_infra_group_min_size: 1'
+            }
```
mikenairn: @odra Does setting these to 1 work, or is it always at least 2?

jobs/openshift/cluster/create/openshift-cluster-create.yaml
```diff
@@ -8,9 +8,12 @@
...
           default: ''
           description: '[REQUIRED] The name of the cluster to create'
-      - string:
+      - choice:
           name: 'awsRegion'
-          default: 'eu-west-2'
+          choices:
+            - 'us-east-1'
+            - 'eu-west-1'
+            - 'eu-west-2'
```
mikenairn: I'm not actually sure selecting anything other than "eu-west-2" works at the moment does it?

jobs/openshift/cluster/create/Jenkinsfile
odra: if it is one it will set to one (it's not working atm) otherwise it will use the  default value

jobs/openshift/cluster/create/openshift-cluster-create.yaml
odra: Just wanted to use the same values used in the destroy job

mikenairn: What's the default value set to if no awsRegion is specified? Can you set the default choice to be "eu-west-2", i'm thinking of jenkins jobs that trigger this one, but might just use the default values.

odra: ok let me set it

mikenairn: I guess this is the only job that i know of that triggers it, and it's specifying the region https://github.com/integr8ly/ci-cd/blob/master/jobs/openshift/cluster/integreatly/test/Jenkinsfile#L30

PR #138, opened 2019-04-16
Intly 1699

pawelpaszki:
## What
https://issues.jboss.org/browse/INTLY-1699

## Verification
Pipeline with same definition was executed here:

https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/ppaszki-pds-patch-install/4/console

To make sure that the versions are verified you can also run smoke tests (there should be only one failure) against this cluster: https://master.ppaszki-7f6a.openshiftworkshop.com 
Additionally, go to the .yaml definition of patched deployments, e.g. grafana-operator

There is also pipeline with patching code running against POC cluster (it does not include the 2 minutes sleep after installation, which IMO is required to prevent from getting 'resource not found' error):

https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/poc-install/45/console

### DIFF
--- a/jobs/integr8ly/pds-install.yaml
+++ b/jobs/integr8ly/pds-install.yaml
@@ -195,6 +195,9 @@
                                 
                                 # patch keycloak-operator to use master tag
                                 sudo oc patch deployment keycloak-operator -n sso -p  '{ "spec": { "template": { "spec": { "containers": [{ "name": "keycloak-operator", "image": "quay.io/integreatly/keycloak-operator:master" }]}}}}'
+
+                                # patch user-sso keycloak-operator to use master tag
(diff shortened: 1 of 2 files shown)


pawelpaszki:
@trepel @tremes mind taking a look?

tremes:
The code is little bit repetitive but it's not a big deal. I would rather see it defined (and it needs to work) in manifest file. Otherwise looks good.

trepel:
@pawelpaszki looking as well

pawelpaszki:
> There is a new component (namespace) called `user-sso`. It is just another Keycloak instance for standard users so the patching process will be the same as for `sso` namespace.

Done.

PR #139, opened 2019-04-17
INTLY-1238 introduce a pipeline for launcher testing

tremes:

### DIFF
--- a/jobs/integr8ly/launcher-test.yaml
+++ b/jobs/integr8ly/launcher-test.yaml
@@ -1,7 +1,7 @@
 ---
 
 - job:
-    name: fuse-tests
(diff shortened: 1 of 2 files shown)


pawelpaszki:
It looks good, however there is one problem - you only see the logs after the execution is completed (after ~ 16 minutes). I would rather vote for defining the test execution stage similarly as here: https://github.com/integr8ly/ci-cd/blob/master/jobs/integr8ly/alerts-test.yaml#L39-L54

tremes:
@pawelpaszki ok thanks. I'll check

tremes:
Updated. I also updated the fuse-test pipeline logging.

PR #140, opened 2019-04-17
added ns-prefix to pds and poc install + uninstall pipelines

pawelpaszki:
## What
https://issues.jboss.org/browse/INTLY-1719

## Verification
PDS uninstall was executed here: https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/ppaszki-pds-ns-prefix-uninstall/1/console
PDS install: https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/ppaszki-pds-ns-prefix-install/2/console

POC install and uninstall have the same changes, so I think it is safe to assume that it will work as well

Note.
This change will require introducing additional parameter to most of our tests and also to nightly executors. I will create those PRs and we would probably need to merge this PR along with the PRs for test suites and changes to nightly executors (which will be made after assessing, which tests need the ns-prefix param)

### DIFF
--- a/jobs/integr8ly/pds-install.yaml
+++ b/jobs/integr8ly/pds-install.yaml
@@ -42,6 +42,10 @@
         - string:
             name: ADDITIONAL_ANSIBLE_PARAMS
             description: "Additional parameters passed to install playbook, e.g.'-e eval_seed_users_count=0'. Can be left empty"
+        - string:
+            name: NAMESPACE_PREFIX
+            default: rhpds-
+            description: "This value will be prefixed to the names of namespaces created during Integr8ly installation"
     dsl: |
         import hudson.model.*
         import jenkins.model.*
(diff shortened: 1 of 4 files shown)


trepel:
the config of nightly triggers can be edited directly in Jenkins. They are not re-created each night by 'recreate-pipelines'. But yes, PR for template update would be better.

tremes:
Looks good, however I don't understand why we use `rhpds` prefix and `openshift` prefix. Doesn't make sense IMO. Both should be `openshift` or `rhpds` and `poc` then...

pawelpaszki:
@trepel can answer that. I don't mind having the same prefix value for both

trepel:
The idea of using different prefixes is that `openshift-` is sort of default prefix (OSD, POC clusters for our customers will have this one), whereas on PDS we can use some different prefix (the actual value does not really matter) in order to catch the errors like hard-coded `openshift-` string in code or not taking ns_prefix into account at all.

Does it make sense?

tremes:
Hm ok thanks

PR #148, opened 2019-04-23
added all tests executor job

pawelpaszki:
## What
https://issues.jboss.org/browse/INTLY-1568

## Verification
The pipeline is being executed here: https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/all-tests-executor/2/console


tremes:
Do we want to run AMQ address creation test and Fuse test and then both of them again within w1 executor test?

pawelpaszki:
I guess we don't. I will leave w1 executor test.

PR #152, opened 2019-04-24
Added namespace prefix to w1A test

trepel:
## What

The AMQ address creation test requires NAMESPACE_PREFIX as input param. The w1A test pipeline triggers the AMQ address creation test pipeline under the hood thus w1A test pipeline must pass this input param to AMQ address creation pipeline.

### DIFF
--- a/jobs/integr8ly/w1-test-executor.yaml
+++ b/jobs/integr8ly/w1-test-executor.yaml
@@ -16,6 +16,9 @@
       - string:
           name: CLUSTER_URL
           description: 'URL of cluster on which the W1 test will be executed.'
+      - string:
+          name: NAMESPACE_PREFIX
+          description: "Value used to prefix the names of the namespaces created during Integr8ly installation"
       - string: 
           name: ADMIN_USERNAME
           default: 'admin@example.com'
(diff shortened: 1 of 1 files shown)


trepel:
Build here:
https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/w1-test-executor/22/console

Still not blue, but at least the AMQ online creation part is ok.

jobs/integr8ly/w1-test-executor.yaml
```diff
@@ -50,6 +53,7 @@
                                     string(name: 'REPOSITORY', value: "${REPOSITORY}"),
                                     string(name: 'BRANCH', value: "${BRANCH}"),
                                     string(name: 'CLUSTER_URL', value: "${CLUSTER_URL}"),
+                                    string(name: 'NAMESPACE_PREFIX', value: "${NAMESPACE_PREFIX}"),
```
tremes: Why do you add it only for AMQ address test?

tremes: hmm I guess we don't use it in Fuse tests yet right?

trepel: Because I thought it was not required for fuse-test. But I am debugging the failure of `fuse-test` and it seems that the namespace prefix needs to be added there as well. However, this will require code changes (MR to GitLab) and then followup PR here. So if you have no objections, I would merge this one and once the MR (I am working on it atm) is merged I can create the followup PR here.

tremes: OK merging

PR #155, opened 2019-04-26
Defaulting to empty NS prefix for RHPDS

trepel:
## What

As per discussion with eng, we only support empty namespace prefix for RHPDS thus the default values for this param were updated to an empty string

## Verification steps
jenkins-jobs --conf jenkins_jobs.ini test jobs/integr8ly/pds-install.yaml
jenkins-jobs --conf jenkins_jobs.ini test jobs/integr8ly/pds-uninstall.yaml
jenkins-jobs --conf jenkins_jobs.ini test jobs/integr8ly/pds-testing-executor.yaml
jenkins-jobs --conf jenkins_jobs.ini test jobs/integr8ly/pds-testing-nightly-trigger.yaml


pawelpaszki:
Maybe we should remove the default alltogether instead of providing empty string?

PR #168, opened 2019-05-02
no patching nightly triggers need to use tests from v1.3

tremes:


pawelpaszki:
Should we merge it now or when the v1.3 is ready?

tremes:
I think we can merge it now. The v1.3 is already in upstream. I tried it with non-patched master...and was thinking that 1.3.0 might be a different story, but it hopefully doesn't matter.

PR #170, opened 2019-05-02
INTLY-1700 add first tests to the Integreatly Test stage

tremes:
~This is WIP. I added first tests to the testing stage and also updated from Jenkins file to the DSL in yaml directly.~ 

~The tests are currently failing (due to AMQ Online update - https://issues.jboss.org/browse/INTLY-1939).~
Please let me know what you think about it.

### DIFF
--- a/jobs/openshift/cluster/integreatly/test/Jenkinsfile
+++ b/jobs/openshift/cluster/integreatly/test/Jenkinsfile
@@ -1,4 +1,5 @@
 def testOptions = [:]
+final QE_REPO_URL = 'https://gitlab.cee.redhat.com/integreatly-qe/integreatly-qe.git'
 
 def verifyParameters(testOptions) {
   if (!testOptions.gitUrl || !testOptions.gitBranch) {
@@ -14,6 +15,18 @@ def verifyParameters(testOptions) {
     verifyParameters(testOptions)
   }
 }
+
+def setClusterAdminCredentials() {
+  def integreatlyCredentialsID = 'tower-openshift-cluster-credentials'
+  def clusterAdminCredentials = [:]
+  withCredentials([usernamePassword(credentialsId: integreatlyCredentialsID, usernameVariable: 'CLUSTER_ADMIN_USERNAME', passwordVariable: 'CLUSTER_ADMIN_PASSWORD')]) {
+    clusterAdminCredentials.clusterAdminUsername = "${CLUSTER_ADMIN_USERNAME}"
+    clusterAdminCredentials.clusterAdminPassword = "${CLUSTER_ADMIN_PASSWORD}"
+  }
+  return clusterAdminCredentials
+}
+
+def clusterAdminCredentials = setClusterAdminCredentials()
 
 pipeline {
   agent {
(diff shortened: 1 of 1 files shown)


tremes:
This first successful (finally :)) run was https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/OpenShift/job/openshift-cluster-integreatly-test-tremes/23/.

@mikenairn @JameelB can you please take a look?

jobs/openshift/cluster/integreatly/openshift-cluster-integreatly-test.yaml
```diff
@@ -17,12 +17,181 @@
...
+                    [$class: 'BooleanParameterValue', name: 'dryRun', value: params.dryRun]
+                  ]
+                  build job: jobName, parameters: jobParams
+                } else {
+                  println "[INFO] Cluster name not available, skipping deprovision"
+                }
+              }
+            }
+          }
+        }
```
mikenairn: Can you revert this and apply all changes in the Jenkinsfile https://github.com/integr8ly/ci-cd/blob/bf4c31b0566c47e04ffcf50f93f287ab09effa72/jobs/openshift/cluster/integreatly/test/Jenkinsfile. We don't put all the pipeline code in the job definition as it's harder to read and edit, and the jobs are eithe rpre processed to inject the dsl like this via the configure_script or loaded from Git.

tremes: Ahh ok. I thought you prefer this. I'll apply the changes in the Jenkinsfile then.

tremes: @mikenairn Updated.

jobs/openshift/cluster/integreatly/test/Jenkinsfile
```diff
@@ -1,4 +1,5 @@
 def testOptions = [:]
+final QE_REPO_URL = 'https://gitlab.cee.redhat.com/integreatly-qe/integreatly-qe.git'
```
mikenairn: Is this repo being versioned now?

jobs/openshift/cluster/integreatly/test/Jenkinsfile
```diff
@@ -68,12 +81,58 @@ pipeline {
...
 
     stage('Test Integreatly') {
       steps {
-        echo "[INFO] Running Tests on Integreatly"
+        echo "[INFO] Running Browser Tests on Integreatly"
+        script { 
+          jobName = 'browser-based-testsuite-pipeline'
+          jobParams = [
+            [$class: 'StringParameterValue', name: 'REPOSITORY', value: QE_REPO_URL],
+            [$class: 'StringParameterValue', name: 'BRANCH', value: 'master'],
```
mikenairn: I'm just wondering here, longer term, should we allow this to be set to match whatever branch you are testing? Or at least check if a branch with the same name exists on this repo, and if it does, use that?

If we are versioning the test repo along with the installer (master -> master, v1.3 -> v1.3) it seems likely that we would want to have this branch change depending on what is actually being tested. It would also allow us to update tests before they hit master in Delorean as we could create for example a 3scale-ga branch on the test repo that can be updated according to changes for the next ga release of 3scale.

jobs/openshift/cluster/integreatly/test/Jenkinsfile
tremes: It is

jobs/openshift/cluster/integreatly/test/Jenkinsfile
tremes: OK it could be possible, So what do you suggest? For example having a dropdown list with available branch names as an input param or something more sophisticated like trying to match the branch name? I would go with some simpler solution first.

mikenairn: Personally if it was me, i'd change the test jobs to check the branch being specified existed and if it doesn't fall back to master, we used to do this for some rhmap stuff see https://github.com/feedhenry/fh-pipeline-library/blob/master/vars/checkoutGitRepo.groovy#L52-L61. The input to all these jobs for BRANCH would then be testOptions.gitBranch, but internally if that didn't exist it would use master.

I don't think doing the drop down initially makes much sense really, as most of this will be automated and we won't know if the branch will exist or not to specify it as an input.

I'm happy enough to leave it as "master" initially as long as we have a follow up jira to make it configurable.

jobs/openshift/cluster/integreatly/test/Jenkinsfile
mikenairn: :+1: I assume this will use the same versioning as the installer? master-> master v1.3 -> v1.3 etc....

tremes: Yes. OK I think this is related to your second comment below right? We will basically try to find out a QE repo branch with the same name as the installation branch. Did I get it correctly?

jobs/openshift/cluster/integreatly/test/Jenkinsfile
tremes: OK thanks for the tips. I think I will create a followup JIRA for this.

tremes: Created https://issues.jboss.org/browse/INTLY-2098

jobs/openshift/cluster/integreatly/test/Jenkinsfile
mikenairn: Yes exactly. We want the versioning to be the same so that we can easily identify which tests should work with which version of integreatly.

tremes:
I created the follow up tasks. See 

- https://issues.jboss.org/browse/INTLY-2098
- https://issues.jboss.org/browse/INTLY-2101
- https://issues.jboss.org/browse/INTLY-2102

Any other thoughts @trepel, @pawelpaszki ?

pawelpaszki:
@tremes I think it makes sense.

PR #171, opened 2019-05-02
Add a pre install stage to the openshift integreatly install pipeline

JameelB:
## What
A pre install stage needs to added to the `Openshift Cluster Integreatly Install` pipeline. This should complete any tasks required before the installation of integreatly is triggered.
- [x] add job for syncing images from brew to the cluster registry to be used for RC testing
- [x] add a preinstall stage to the installation pipeline
    - Trigger a build for `Openshift Cluster Integreatly Brew Images Sync` from this stage to ensure that RC images will be available during installation

Image sync config file format:
```
imageSync:
  sourceRepo: <default-source-repo>
  projects:
    <project-name>:
      images:
      - [images]
    <project-name>:
      sourceRepo: <source-repo>
      images:
      - [images]
...
```

### DIFF
--- a/jobs/openshift/cluster/brew/Jenkinsfile
+++ b/jobs/openshift/cluster/brew/Jenkinsfile
@@ -55,6 +55,7 @@ stage('Image Sync Options') {
 
 def githubSSHCredentialsID = 'jenkinsgithub'
 def imageSyncConfigFile = './integreatly.yml'
+def imageSyncConfigContent
 String brewContainerRegistry
 def projectsToSync
 def syncImages = false
@@ -64,21 +65,21 @@ node('cirhos_rhel7') {
     stage('Ensure Sync File Exists') {
         dir('installation') {
             gitCheckoutRepo(imageSyncOptions.installationGitUrl, imageSyncOptions.installationGitBranch, githubSSHCredentialsID, '.')
-            syncImages = fileExists imageSyncConfigFile
+            def imageSyncFileExists = fileExists imageSyncConfigFile
+            imageSyncConfigContent = readYaml file: imageSyncConfigFile
 
-            if (!syncImages) {
-                println "Image sync file '${imageSyncConfigFile}' is not available. No images to sync, skipping!"
+            if (imageSyncFileExists && imageSyncConfigContent.imageSync && imageSyncConfigContent.imageSync.projects) {
+                syncImages = true
+            } else {
+                println "No images to sync, skipping!"
             }
         }
     }
 
     stage('Get Images To Sync') {
         when(syncImages) {
-            dir('installation') {
-                def imageSyncConfigContent = readYaml file: imageSyncConfigFile
-                projectsToSync = imageSyncConfigContent.imageSync
-                brewContainerRegistry = imageSyncConfigContent.sourceRepo
-            }
+            projectsToSync = imageSyncConfigContent.imageSync.projects
+            brewContainerRegistry = imageSyncConfigContent.imageSync.sourceRepo
         }
     }
 
@@ -114,7 +115,7 @@ node('cirhos_rhel7') {
                 brewContainerRegistry = projectConfig.sourceRepo ?: brewContainerRegistry
 
                 if (!brewContainerRegistry) {
-                  error "Brew container registry was not specified"
+                    error "Brew container registry was not specified"
                 } else {
                     projectConfig.images.each {
                         def brewImageUrl = "${brewContainerRegistry}/${project}/${it}"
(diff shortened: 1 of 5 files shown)


jobs/openshift/cluster/brew/openshift-cluster-brew-image-sync-1835.yaml
```diff
@@ -0,0 +1,40 @@
+---
+- job:
+    name: openshift-cluster-brew-image-sync-1835
```
JameelB: NOTE: Will need to remove `-1835` from the job template name when this job is ready to be used.

jobs/openshift/cluster/brew/Jenkinsfile
```diff
@@ -0,0 +1,143 @@
...
+String brewContainerRegistry
+def projectsToSync
+def syncImages = false
+
+node('cirhos_rhel7') {
+    cleanWs()
+    stage('Ensure Sync File Exists') {
+        dir('installation') {
+            gitCheckoutRepo(imageSyncOptions.installationGitUrl, imageSyncOptions.installationGitBranch, githubSSHCredentialsID, '.')
+            syncImages = fileExists imageSyncConfigFile
```
mikenairn: This is setting syncImages to true if the file "integreatly.yml" exists, but should it not also check to see if there is a "syncImages" section in the file?

jobs/openshift/cluster/brew/Jenkinsfile
```diff
@@ -0,0 +1,143 @@
...
+                println "Image sync file '${imageSyncConfigFile}' is not available. No images to sync, skipping!"
+            }
+        }
+    }
+
+    stage('Get Images To Sync') {
+        when(syncImages) {
+            dir('installation') {
+                def imageSyncConfigContent = readYaml file: imageSyncConfigFile
+                projectsToSync = imageSyncConfigContent.imageSync
```
mikenairn: This is assuming "imageSync" exists in the file, see comment above about checking for it's existence when setting "syncImages".

jobs/openshift/cluster/brew/Jenkinsfile
```diff
@@ -0,0 +1,143 @@
...
+                brewContainerRegistry = imageSyncConfigContent.sourceRepo
```
mikenairn: This should allow a sourceRepo to be set at the higher level right?

```

imageSync:
  sourceRepo: brew-pulp-docker01.web.prod.ext.phx2.redhat.com:8888
  fuse7:
    images:
    - fuse-ignite-meta:1.3-16
```

jobs/openshift/cluster/brew/Jenkinsfile
```diff
@@ -0,0 +1,143 @@
...
+        when(syncImages) {
+            String clusterContainerRegistryConsole = "registry-console-default.apps.${getHost(imageSyncOptions.openshiftMasterUrl)}" 
+            projectsToSync.keySet().each {
+                if (params.dryRun) {
+                    println "Would create a new project called '${it}' and set the registry project access policy to 'Anonymous: Allow all unauthenticated users to pull images'"
+                } else {
+                    sh "oc new-project ${it} || true"
+                    // Set registry project access policy to Anonymous: Allow all unauthenticated users to pull images
+                    sh "oc create rolebinding registry-viewer --clusterrole=registry-viewer || true"
+                    sh "oc adm policy add-role-to-group registry-viewer system:authenticated system:unauthenticated || true"
```
mikenairn: Cool, so this will allow any user/project to pull images from any of these registries we are syncing right?

jobs/openshift/cluster/brew/openshift-cluster-brew-image-sync-1835.yaml
mikenairn: I think you can just remove it now and update the real job. I thought it was already updated tbh.

jobs/openshift/cluster/integreatly/install/Jenkinsfile
```diff
@@ -59,6 +59,20 @@ stage("Verify Installation Options") {
...
+      def jobParams = [
+              [$class: 'StringParameterValue', name: 'openshiftMasterUrl', value: installerOptions.openshiftMasterUrl],
+              [$class: 'StringParameterValue', name: 'clusterAdminUsername', value: installerOptions.clusterAdminUsername],
+              [$class: 'StringParameterValue', name: 'clusterAdminPassword', value: installerOptions.clusterAdminPassword],
+              [$class: 'StringParameterValue', name: 'installationGitUrl', value: installerOptions.installationGitUrl],
+              [$class: 'StringParameterValue', name: 'installationGitBranch', value: installerOptions.installationGitBranch],
+              [$class: 'BooleanParameterValue', name: 'dryRun', value: params.dryRun]
+      ]
+      build job: jobName, parameters: jobParams
+    }
```
mikenairn: Can you move this stage outside of the node('cirhos_rhel7'). Doing it this way will block 2 executors since the job being called is also requesting a node.

jobs/openshift/cluster/brew/Jenkinsfile
JameelB: yes this should allow that. If you go into the registry project, it should show that the access policy is set to `Anonymous: Allow all unauthenticated users to pull images`

jobs/openshift/cluster/brew/Jenkinsfile
JameelB: I was thinking it would be on the same level as `imageSync`. That would make it easier to iterate through the list of projects without having to ignore `sourceRepo`. But if there would be other stuff in this file apart form imageSync, then it makes sense to have that in there.

PR #177, opened 2019-05-07
patching webapp operator to master

pawelpaszki:
## What
patching webapp operator to master

## Verification
run the install pipelines with PATCH_TO_MASTER=true and confirm that webapp operator image tag equals master. Alternatively check this pipeline build output: https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/pds-install/194/console


tremes:
@pawelpaszki scale down and up is not required in this case? Otherwise LGTM.

pawelpaszki:
it is required. Not really a part of this PR, but tested with the other PR and without the scale down/ up there were 2 pods of the operators. The new one was not functioning because of the first one not being deleted

pawelpaszki:
@tremes @trepel can you please have a look?

PR #178, opened 2019-05-07
implement get latest prerelease for 3scale, fuse online and amq

JameelB:
## What
- [x] Create a function to get the latest prerelease of 3Scale
- [x] Create a function to get the latest prerelease of Fuse Online
- [x] Create a function to get the latest prerelease of AMQ Online

To be verified with: https://github.com/integr8ly/delorean-pipeline-library/pull/5

### DIFF
--- a/jobs/delorean/jenkinsfiles/discovery/rc/github/Jenkinsfile
+++ b/jobs/delorean/jenkinsfiles/discovery/rc/github/Jenkinsfile
@@ -70,18 +70,43 @@ String getProductVersionID(productName) {
     return result
 }
 
-String getLatestPreRelease(projectOrg, projectRepo, githubToken, productName) {
-    //ToDo Implement this to return the latest pre release (RC) that is available, if any
+String getThreescalePrerelease(projectOrg, projectRepo, githubToken) {
+    def ghUtils = new org.integr8ly.GitHubUtils()
+    def rcTagRef = '-CR'
+    return ghUtils.ghGetLatestReleaseByTag(projectOrg, projectRepo, githubToken, rcTagRef)
+}
+
+String getAmqOnlinePrerelease(projectOrg, projectRepo, githubCredentialsID) {
+    def amqOnlineGitUrl = "git@github.com:${projectOrg}/${projectRepo}.git"
+    def amqOnlineMasterBranch = "amq-online-1-dev"
+    def amqOnlineLatestRC = ""
+    def rcFilter = "grep 'amq-online-' | grep -v 'dev'"
+    
+    dir('amq_online') {
+        gitCheckoutRepo(amqOnlineGitUrl, amqOnlineMasterBranch, githubCredentialsID, '.')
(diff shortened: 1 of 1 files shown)


jobs/delorean/jenkinsfiles/discovery/rc/github/Jenkinsfile
```diff
@@ -70,18 +70,43 @@ String getProductVersionID(productName) {
...
+    def amqOnlineGitUrl = "git@github.com:${projectOrg}/${projectRepo}.git"
+    def amqOnlineMasterBranch = "amq-online-1-dev"
+    def amqOnlineLatestRC = ""
+    def rcFilter = "grep 'amq-online-' | grep -v 'dev'"
+    
+    dir('amq_online') {
+        gitCheckoutRepo(amqOnlineGitUrl, amqOnlineMasterBranch, githubCredentialsID, '.')
+        amqOnlineLatestRC = sh(returnStdout: true, script: "git for-each-ref refs/remotes/ --sort=-creatordate --format='%(refname:short)' | ${rcFilter} | head -n 1").trim()
+        amqOnlineLatestRC = amqOnlineLatestRC.minus("origin/")
+    }
```
mikenairn: I'm always a bit sceptical of using the date something was created to determine something as being a more recent version.

Since we know the branch pattern they have is "amq-online-<major><minor>", could we not just parse the branch names and use the last part of it to determine the version? Assuming the next one will be amq-online-12, it will be greater than amq-online-11?

jobs/delorean/jenkinsfiles/discovery/rc/github/Jenkinsfile
```diff
@@ -70,18 +70,43 @@ String getProductVersionID(productName) {
...
+
+    return amqOnlineLatestRC
+}
+
+String getFuseOnlinePrerelease(projectOrg, projectRepo, githubToken) {
+    def ghUtils = new org.integr8ly.GitHubUtils()
+    return ghUtils.ghGetLatestReleaseByTag(projectOrg, projectRepo, githubToken)
```
mikenairn: Not important to change now, but might be worth considering making this method static in the library

jobs/delorean/jenkinsfiles/discovery/rc/github/Jenkinsfile
JameelB: I believe that creatordate actually looks at the commit date. I was thinking since it's listed alphabetically, when it comes round to having a major `10+`, assuming itll be formatted as amq-online-10x, that would be placed between `1x` and `2x`. But that's probably far into the future, i'm ok with dropping the sort by.

mikenairn: I see your point about the major minor when they get past a certain point since they don't put any delimiter between the major and minor versions.

I'd image the creatordate is definitely using the commit date, not sure what else it would be using, but i can't see how you can use that as a guarantee that this was actually made in relation to the latest release, could just as easily be being made on a branch for an older version that is still being maintained.

Both approaches have issues.

mikenairn: I'd say lets go with what you have for now, we can iterate on it later if we come up with a better approach.

PR #185, opened 2019-05-17
Do not run anything on the Jenkins master node

tremes:


tremes:
@mikenairn Can you check as well please?

mikenairn:
I don't see any issues with it, but really if you don't care what node gets used, you shouldn't need to specify. Maybe you should just disable job execution of the master?

tremes:
Yes good point. I changed master node to "Only build jobs with label expressions matching this node". So this could be sufficient.

trepel:
@tremes cool, are we ok to close this one without merging? I would say yes, or would you like to merge anyway?

tremes:
Nah...better to close

PR #187, opened 2019-05-17
Increased timeout to 4:30 h for full testing

trepel:
## What

This build (https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/pds-testing-executor/104/console) seems to be timeouted. As we keep adding more and more tests it seems we need to increase timeout (3:30h are not enough). Thus this PR increases the timeout by 1 hour to 4:30h (270 min).


trepel:
Not sure if there is anything to do to verify here.

PR #190, opened 2019-05-24
INTLY-847 introduce a MSB integration test pipeline with help of Ans…

tremes:
It's available at https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/msb-integration-test/

The managed-service-broker integration tests are currently failing so I will try to investigate and fix it.

### DIFF
--- a/jobs/integr8ly/msb-integration-test.yaml
+++ b/jobs/integr8ly/msb-integration-test.yaml
@@ -1,0 +1,184 @@
+---
+- job:
+    name: msb-integration-test
+    display-name: 'Managed Service Broker Integration Test'
+    project-type: pipeline
+    concurrent: false
+    triggers:
+      - timed: "15 01 * * *"
+    properties:
+      - build-discarder:
+          num-to-keep: 15
+    parameters:
+      - string:
+          name: 'installationGitUrl'
+          default: 'https://github.com/integr8ly/installation.git'
+          description: '[REQUIRED] Integreatly installer Git URL'
+      - string:
+          name: 'installationGitBranch'
+          default: 'master'
+          description: '[REQUIRED] The name of the git branch to be used for installing Integreatly'
+      - string:
+          name: RECIPIENTS
+          default: integreatly-qe@redhat.com
+          description: "Whitespace- or comma-separated list of recipient addresses"
+    dsl: |
+        def testOptions = [:]
+        def error = null
+
+        def verifyParameters(testOptions) {
+          if (!testOptions.gitUrl || !testOptions.gitBranch) {
+            def userInput = input message: 'Integreatly Test Options', parameters: [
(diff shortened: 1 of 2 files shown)


jobs/integr8ly/msb-integration-test.yaml
```diff
@@ -0,0 +1,162 @@
...
+              stage('Verify Parameters') {
+                testOptions.awsAccountName = 'fheng.AWS'
+                testOptions.awsRegion = 'eu-west-2'
+                testOptions.clusterName = params.installationGitBranch
+                testOptions.domainName = 'skunkhenry.com'
+                testOptions.gitUrl = params.installationGitUrl
+                testOptions.gitBranch = params.installationGitBranch
+                testOptions.openshiftMasterUrl = "https://${params.installationGitBranch}.${testOptions.domainName}"
+                verifyParameters(testOptions)
+              }
```
tremes: We can remove this stage in the future or just change it in default params initialization.

tremes:
Maybe we can define this as nightly pipeline by default. WDYT?

trepel:
> Maybe we can define this as nightly pipeline by default. WDYT?

This is definitely a candidate for nightly execution. You can create a nightly trigger or just schedule this job to run each night - I am ok either way.

tremes:
@trepel yes I was thinking about the second option. I think having an extra nightly trigger is somewhat unnecessary

tremes:
It runs each night at 01:15 am here https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/Nightly/job/msb-integration-test/. 

Now it should also send a notification email and finally the fix for the test is at https://github.com/integr8ly/managed-service-broker/pull/41

jobs/integr8ly/msb-integration-test.yaml
```diff
@@ -0,0 +1,181 @@
...
+          name: 'installationGitUrl'
+          default: 'https://github.com/integr8ly/installation.git'
+          description: '[REQUIRED] Integreatly installer Git URL'
+      - string:
+          name: 'installationGitBranch'
+          default: 'master'
+          description: '[REQUIRED] The name of the git branch to be used for installing Integreatly'
+    dsl: |
+        def testOptions = [:]
+        final QE_REPO_URL = 'https://gitlab.cee.redhat.com/integreatly-qe/integreatly-qe.git'
```
trepel: Is this `QE_REPO_URL` used anywhere?

To be honest I am not sure if these lines of code around `testOptions` are indeed required here. It is part of the Cluster Create and Integreatly Install jobs anyway.  But I am not familiar with the stuff enough so it might be good having it here.

tremes: It's not. It can be removed.

tremes:
Hopefully last update - adding RECIPIENTS variable.

tremes:
@trepel Finally first blue result. :) merging

PR #191, opened 2019-05-27
use github user/pass creds for authenticating github API requests

JameelB:
## What
Use Github user/password credentials for authenticating github API requests instead of the token.

To be tested with: https://github.com/integr8ly/delorean-pipeline-library/pull/8

### DIFF
--- a/jobs/delorean/jenkinsfiles/discovery/ga/github/Jenkinsfile
+++ b/jobs/delorean/jenkinsfiles/discovery/ga/github/Jenkinsfile
@@ -3,8 +3,8 @@
 
 def installationGitUrl = params.installationGitUrl ?: 'git@github.com:integr8ly/installation.git'
 def installationGitRef = params.installationGitRef ?: 'master'
-def githubToken = params.githubToken ?: 'jenkins-github-api-token'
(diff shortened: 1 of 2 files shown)


jobs/delorean/jenkinsfiles/discovery/ga/github/Jenkinsfile
```diff
@@ -3,8 +3,8 @@
 
 def installationGitUrl = params.installationGitUrl ?: 'git@github.com:integr8ly/installation.git'
 def installationGitRef = params.installationGitRef ?: 'master'
-def githubToken = params.githubToken ?: 'jenkins-github-api-token'
-def githubCredentialsID = params.credentialId ?: 'jenkinsgithub'
+def githubCredentialsID = params.credentialId ?: 'githubjenkins'
+def githubSSHCredentialsID = params.sshCredentialId ?: 'jenkinsgithub'
```
mikenairn: I assume none of these were ever coming in as params and we were always suing the default fallback variable?

JameelB: Yeah the default vars were always being used. We don't have this param set up in the jobs itself.

PR #195, opened 2019-05-30
INTLY-2102 execute all-tests-executor within the Delorean pipeline

tremes:
See the run https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/OpenShift/job/openshift-cluster-integreatly-test-tremes/24 which was pretty good (only smoke tests failed - I'll investigate it)

### DIFF
--- a/jobs/integr8ly/all-tests-executor.yaml
+++ b/jobs/integr8ly/all-tests-executor.yaml
@@ -82,7 +82,7 @@
           default: '1'
           description: 'optionally increase timeout values. If you provide value 2 it means it will be sleeping/waiting two times longer'      
     dsl: |
-        parameters = [
+        jobParams = [
             string(name: 'REPOSITORY', value: "${REPOSITORY}"),
             string(name: 'BRANCH', value: "${BRANCH}"),
             string(name: 'CLUSTER_URL', value: "${CLUSTER_URL}"),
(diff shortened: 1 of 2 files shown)


jobs/integr8ly/all-tests-executor.yaml
```diff
@@ -150,7 +150,7 @@
         }
 
         def runTests(testPipeline) {
-            buildStatus = build(job: testPipeline, propagate: false, parameters: parameters).result
+            buildStatus = build(job: testPipeline, propagate: false, parameters: jobParams).result
             println "Build finished with ${buildStatus}"
```
tremes: These both changes are not necessary (just leftover after my experiments) on the other hand this is little bit more clear.

pawelpaszki:
> https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/OpenShift/job/openshift-cluster-integreatly-test-tremes/24

I think smoke tests failed, because there is no Gitea NS

jobs/integr8ly/all-tests-executor.yaml
pawelpaszki: I don't mind either way

tremes:
@pawelpaszki yeah I know :)

pawelpaszki:
@tremes I think it is deliberately not installed on OSD as well as a couple of other namespaces. Anyway - LGTM

tremes:
@pawelpaszki yes I think you're right. I'll check. Thanks for review anyway.

trepel:
@tremes Yes, some namespaces are not installed on OSD on purpose. We should be tracking which ones in our test plan.

PR #202, opened 2019-06-06
INTLY-2273 use different cluster name than "master" in Delorean testi…

tremes:
…ng pipeline.

Running at https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/OpenShift/job/openshift-cluster-integreatly-test-tremes/25 so let's see.

### DIFF
--- a/jobs/openshift/cluster/integreatly/openshift-cluster-integreatly-test.yaml
+++ b/jobs/openshift/cluster/integreatly/openshift-cluster-integreatly-test.yaml
@@ -13,6 +13,9 @@
           name: 'installationGitBranch'
           default: 'master'
           description: '[REQUIRED] The name of the git branch to be used for installing Integreatly'
+      - string:
+          name: 'clusterName'
+          description: '[OPTIONAL] The name of the newly provisioned cluster on skunkhenry domain. Defaults to branch name if not provided.'    
       - bool:
           name: 'dryRun'
           default: false
--- a/jobs/openshift/cluster/integreatly/test/Jenkinsfile
+++ b/jobs/openshift/cluster/integreatly/test/Jenkinsfile
@@ -10,8 +10,10 @@ def verifyParameters(testOptions) {
 
     testOptions.gitUrl = userInput.installationGitUrl
     testOptions.gitBranch = userInput.installationGitBranch
-    testOptions.clusterName = userInput.installationGitBranch
-    testOptions.openshiftMasterUrl = "https://${userInput.installationGitBranch}.${testOptions.domainName}"
+    if(!testOptions.clusterName) {
+      testOptions.clusterName = userInput.installationGitBranch
+      testOptions.openshiftMasterUrl = "https://${userInput.installationGitBranch}.${testOptions.domainName}"
+    }
     verifyParameters(testOptions)
   }
 }
@@ -55,11 +57,11 @@ pipeline {
         script {
           testOptions.awsAccountName = 'fheng.AWS'
           testOptions.awsRegion = 'eu-west-2'
-          testOptions.clusterName = params.installationGitBranch
+          testOptions.clusterName = params.clusterName ? params.clusterName : params.installationGitBranch
           testOptions.domainName = 'skunkhenry.com'
           testOptions.gitUrl = params.installationGitUrl
           testOptions.gitBranch = params.installationGitBranch
-          testOptions.openshiftMasterUrl = "https://${params.installationGitBranch}.${testOptions.domainName}"
+          testOptions.openshiftMasterUrl = "https://${testOptions.clusterName}.${testOptions.domainName}"
           verifyParameters(testOptions)
         }
       }


tremes:
@mikenairn I was just wondering why we call the verification methods two times ? I mean e.g 
https://github.com/integr8ly/ci-cd/blob/master/jobs/openshift/cluster/integreatly/install/Jenkinsfile#L24 and 
https://github.com/integr8ly/ci-cd/blob/master/jobs/openshift/cluster/integreatly/install/Jenkinsfile#L51

https://github.com/integr8ly/ci-cd/blob/master/jobs/openshift/cluster/integreatly/test/Jenkinsfile#L15 and
https://github.com/integr8ly/ci-cd/blob/master/jobs/openshift/cluster/integreatly/test/Jenkinsfile#L63

It seems to me weird but maybe I am missing something since this pattern appears in every Jenkinsfile (within this openshift dir)

tremes:
Hm maybe it will be better to extract the cluster name to a new input param. I think I'll update it.

jobs/openshift/cluster/integreatly/openshift-cluster-integreatly-test.yaml
```diff
@@ -13,6 +13,10 @@
           name: 'installationGitBranch'
           default: 'master'
           description: '[REQUIRED] The name of the git branch to be used for installing Integreatly'
+      - string:
+          name: 'clusterName'
+          default: 'qe-master'
+          description: '[REQUIRED] The name of the newly provisioned cluster on skunkhenry domain.'
```
mikenairn: What's the reason for adding this, is it due to some cert issue or ......?

jobs/openshift/cluster/integreatly/test/Jenkinsfile
```diff
@@ -10,9 +10,6 @@ def verifyParameters(testOptions) {
 
     testOptions.gitUrl = userInput.installationGitUrl
     testOptions.gitBranch = userInput.installationGitBranch
-    testOptions.clusterName = userInput.installationGitBranch
-    testOptions.openshiftMasterUrl = "https://${userInput.installationGitBranch}.${testOptions.domainName}"
-    verifyParameters(testOptions)
```
mikenairn: This should be added back "verifyParameters". It's called from within itself to ensure that whatever the user inputs is also verified. It will just constantly ask for input until such time as the user has entered valid input. As you have it now, someone could just enter nothing and press continue and it would carry on but with invalid inputs and fail at some point later on.

jobs/openshift/cluster/integreatly/openshift-cluster-integreatly-test.yaml
tremes: Well we want to run more things nightly (e.g the MSB integration tests) so we think it's better to use a different name than "master" because having "master" everywhere could mean pipeline clashing problems in the future.

mikenairn: Can you add some details of why we are doing this to the PR description, link to Jira maybe?

There are issues with using the branch name like we are and we do need to look into using a different naming scheme by default, but the changes you have made to this will cause every single cluster to be called "qe-master" unless you override it in every job that is calling it.

A better option for this would be to leave it using the branch name for now, but check the clusterName param and set it to that value if it's not blank, that way everything will still work as normal but allow you to override it in your nightly jobs.

Do you have details of what exactly you are planning re. nightly jobs? I was under the impression you would have a single nightly job that ran all tests, so not sure where this issue came from about needing several clusters at once.

jobs/openshift/cluster/integreatly/test/Jenkinsfile
tremes: OK I didn't understand that. I'll put it back.

jobs/openshift/cluster/integreatly/openshift-cluster-integreatly-test.yaml
tremes: JIRA is https://issues.jboss.org/browse/INTLY-2273. I am failing to see the difference in name "master" vs "qe-master". I am not changing anything in branch naming AFAIK. 

Yes this is our primary pipeline (running all the tests). So maybe we can just change the clusterName in the MSB tests pipeline and keep this as is. 
I am not aware of any further plans for additional nightly pipelines (using tower tooling).

trepel: +1 Using branch as clusterName is not an optimal approach and needs some adjustments
+1 not having `qe-admin` hard-coded as default here - it should only be used in one of our nightly pipelines, not in this job in general

There will be more nightly jobs than just one single nightly job (MSB integration test was mentioned as example). Also, we want to use this for testing per PR potentially. Each pipeline that (eventually) creates cluster should have cluserName as input parameter. I am ok with defaulting to branch name. So we can specify some clusterName that is unlikely to be used by somebody else in our nightly pipeline.

tremes: OK I tried to update it accordingly. Need to check if it works...

trepel: @mikenairn do you have any concerns about this PR?

PR #205, opened 2019-06-07
ensure all-tests-executor job does not run when dryRun is set to true

JameelB:
## What
When `dryRun` is set to true, the cluster create/install workflow in Ansible Tower are not triggered. This only prints what this would do. We need to ensure that the tests are not ran when this parameter is true as it would always fail otherwise due to missing cluster and Integreatly install. 

- [x] Check if `dryRun` is set to true before triggering build for the job `all-tests-executor`.


tremes:
Yes I think this is OK.Thanks

PR #206, opened 2019-06-07
Update default value of master host in inventory template

matskiv:
## What
Updating the value based on this PR: https://github.com/integr8ly/installation/pull/677

Relevant code change: https://github.com/integr8ly/installation/pull/677/files#diff-671124057910da5bb44047033c60d55aL16


tremes:
I am fine with that if it works :-)

matskiv:
Now that I think about it - it's not compatible with 1.4 release. Let me fix that.

trepel:
@matskiv good point. Since we support N and N-1 versions in RHPDS, we need to be able to install both versions there. If the installation process is different for each, we need to update the RHPDS system to be able to install both versions.

matskiv:
I fixed some mistakes I made. This is ready for review now.

PR #207, opened 2019-06-10
INTLY-2266 - remove unused RELASE_VERSION parameter from testing pipe…

tremes:
…lines

### DIFF
--- a/jobs/integr8ly/after-first-login-tests.yaml
+++ b/jobs/integr8ly/after-first-login-tests.yaml
@@ -26,10 +26,6 @@
           default: 'Password1'
           description: 'Evals user password'    
       - string:
-          name: RELEASE_VERSION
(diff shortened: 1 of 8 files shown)


pawelpaszki:
@tremes can you please update pipeline definitions triggers after merging?

tremes:
@pawelpaszki I don't think it' so urgent :) They are updated each night automatically. Anyway thanks for checking.

trepel:
Merged, I manually removed the RELEASE_VERSION from both pds nightly triggers since they are not updated each night automatically. fyi @tonyxrmdavidson - if there is some failure/error speaking about RELEASE_VERSION, it might be caused by my manual modifications.

tremes:
Ahh yeah good point. Sorry and thanks :)

PR #209, opened 2019-06-12
Add new pipeline for codeready test

tremes:

### DIFF
--- a/jobs/integr8ly/codeready-test.yaml
+++ b/jobs/integr8ly/codeready-test.yaml
@@ -1,0 +1,84 @@
+---
+
+- job:
+    name: codeready-tests
+    project-type: pipeline
+    sandbox: true
+    parameters:
+      - string:
+          name: REPOSITORY
+          default: 'https://gitlab.cee.redhat.com/integreatly-qe/integreatly-qe.git'
+          description: 'QE Git repository URL.'
+      - string:
+          name: BRANCH
+          default: 'master'
+          description: 'Branch of the repository'
+      - string:
+          name: CLUSTER_URL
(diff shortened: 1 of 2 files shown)


jobs/integr8ly/codeready-test.yaml
```diff
@@ -0,0 +1,84 @@
...
+          default: 'admin@example.com'
+          description: 'User name to login to Integreatly cluster.'
+      - string: 
+          name: ADMIN_PASSWORD
+          default: 'Password1'
+          description: 'Password to login to Integreatly cluster.'
+      - string:
+          name: EVALS_USERNAME
+          default: 'evals11@example.com'
+          description: 'Evals account email address used to access the available consoles. This test expects that this user started the Walkthrough 1A!!'
```
trepel: Is it really true that the user needs to start wt1A for this test?

tremes: No :)

jobs/integr8ly/codeready-test.yaml
```diff
@@ -0,0 +1,84 @@
...
+      - bool:
+          name: CLEAN_RESOURCES
+          default: true
+          description: 'depending on whether the resources should be cleaned after the successful execution'
+      - string:
+          name: TIMEOUT_THRESHOLD
+          default: '1'
+          description: 'optionally increase timeout values. If you provide value 2 it means it will be sleeping/waiting two times longer'      
+    dsl: |
+        timeout(60) { ansiColor('gnome-terminal') { timestamps {
```
pawelpaszki: probably not major, but timeout here could be significantly reduced, since the text execution takes much less

PR #214, opened 2019-06-20
Add plugin install role

JameelB:
## What
Add role for installing plugins required by Delorean jobs.

- [x] Add role for installing plugins
- [x] Add task to the `configure_jenkins` file to install plugins
- [x] Create `plugins.txt` with a list of plugins to be installed

## Verification
1. Run the script
2. Ensure all plugins are installed.


scripts/plugins.txt
```diff
@@ -0,0 +1,22 @@
...
+gitlab-plugin:1.5.9
+gitlab-hook:1.4.2
+ghprb:1.42.0
+github-branch-source:2.3.6
+extended-choice-parameter:0.76
+delivery-pipeline-plugin:1.2.0
+cvs:2.14
+aws-sqs:2.0.1
+ansicolor:0.5.2
+ansible-tower:0.9.1
```
JameelB: @mikenairn @trepel Are all the required plugins included in this file? Are there any here that are not needed and can be removed?

mikenairn: @JameelB Currently installed plugins are https://gist.github.com/mikenairn/acce9a8150c3d864fce046b1b150a0fd

You won't need to list all of them in here as a lot will be installed as part of Jenkins, but it might give you an idea of some that we rely on that might be missing from this list

trepel: I don't really know what plugins are required for the tests, sorry. I can remember us installing a few, but nothing crazy. I think the best way to proceed is to just proceed with what is here now and if it turns out the test suites require some other plugin, we will update the list.

mikenairn: Did a quick check and these seem to be the plugins we have installed that aren't installed as part of jenkins:

```
< ansible-tower
< ant
< aws-sqs
< build-pipeline-plugin
< cloud-stats
< command-launcher
< cvs
< delivery-pipeline-plugin
< email-ext
< embeddable-build-status
< external-monitor-job
< ghprb
< gitlab-hook
< gitlab-plugin
< groovy
< jms-messaging
< junit-attachments
< kerberos-sso
< ldap
< nodejs
< openstack-cloud
< pam-auth
< redhat-ci-plugin
< role-strategy
< ruby-runtime
< saferestart
< thinBackup
< timestamper
< windows-slaves
```

I think you have most, if not all covered, we can obviously add any as we find them.

PR #215, opened 2019-06-21
Add preflight checks to ansible playbook

JameelB:
## What
Create a role for running preflight checks before proceeding to configure Jenkins. These should check the following:
- [x] Ensure that the credentials to be created are not available in the Jenkins instance.
    - If they are, prompt the user if they want to continue as these credentials will be overwritten
- [x] Ensure there are no plugin incompatibilities.

### DIFF
--- a/scripts/playbooks/roles/preflight_checks/tasks/check_plugins.yaml
+++ b/scripts/playbooks/roles/preflight_checks/tasks/check_plugins.yaml
@@ -29,7 +29,7 @@
   - name: Get list of incompatible plugins
     set_fact:
       incompatible_plugins: "{{ incompatible_plugins | combine({plugin.key: {'current_version': installed_plugins[plugin.key], 'target_version': plugin.value }}) }}"
-    when: installed_plugins[plugin.key] is defined and installed_plugins[plugin.key] != "{{ plugin.value | replace(' ', '') }}"
+    when: installed_plugins[plugin.key] is defined and installed_plugins[plugin.key] != plugin.value | replace(' ', '')
     with_dict: "{{ plugins_to_install }}"
     loop_control:
       loop_var: "plugin"
--- a/scripts/playbooks/roles/preflight_checks/tasks/check_plugins.yaml
+++ b/scripts/playbooks/roles/preflight_checks/tasks/check_plugins.yaml
@@ -35,6 +35,11 @@
       loop_var: "plugin"
   when: get_installed_plugins.stdout_lines | length > 0
 
-- fail:
-    msg: "Found the following incompatible plugins, please update these to the target version. {{ incompatible_plugins }}"
+- pause:
+    prompt: "Incompatible plugins found. This may cause the Delorean jobs to not work as expected.\n{{ incompatible_plugins }}\nDo you want to continue? [y/n]"
+  register: user_confirmation
   when: incompatible_plugins != {}
+
+- fail:
+    msg: "Jenkins configuration discontinued. Please update or uninstall these plugins before running this script again."
+  when: user_confirmation.user_input is defined and user_confirmation.user_input != 'y'
--- a/scripts/playbooks/roles/preflight_checks/tasks/check_plugins.yaml
+++ b/scripts/playbooks/roles/preflight_checks/tasks/check_plugins.yaml
@@ -1,0 +1,40 @@
+---
+- set_fact:
+    base_dir: "{{ playbook_dir | dirname}}"
+
+# Create config file for jenkins plugin install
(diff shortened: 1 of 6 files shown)


scripts/playbooks/roles/preflight_checks/tasks/check_plugins.yaml
```diff
@@ -0,0 +1,40 @@
...
+      installed_plugins: "{{ installed_plugins | combine({item.split(':')[0]: item.split(':')[1]}) }}"
+    with_items: "{{ get_installed_plugins.stdout_lines}}"
+  - name: Parse list of plugins to install
+    set_fact: 
+      plugins_to_install: "{{ plugins_to_install | combine({item.split(':')[0]: item.split(':')[1]}) }}"
+    with_items: "{{ lookup('file', '{{ base_dir }}/plugins.txt').splitlines() }}"
+  - name: Get list of incompatible plugins
+    set_fact:
+      incompatible_plugins: "{{ incompatible_plugins | combine({plugin.key: {'current_version': installed_plugins[plugin.key], 'target_version': plugin.value }}) }}"
+    when: installed_plugins[plugin.key] is defined and installed_plugins[plugin.key] != "{{ plugin.value | replace(' ', '') }}"
```
mikenairn: This seems to be causing this warning:

```[WARNING]: conditional statements should not include jinja2 templating delimiters such as {{ }} or {% %}. Found: installed_plugins[plugin.key] is defined and installed_plugins[plugin.key] != "{{ plugin.value | replace(' ', '') }}" ```

Think you can get rid of the "{{ }}" around plugin.value | replace(' ', '')

File: scripts/playbooks/roles/preflight_checks/tasks/check_plugins.yaml [hunk 2]
```diff
@@ -0,0 +1,40 @@
...
+    with_dict: "{{ plugins_to_install }}"
+    loop_control:
+      loop_var: "plugin"
+  when: get_installed_plugins.stdout_lines | length > 0
+
+- fail:
+    msg: "Found the following incompatible plugins, please update these to the target version. {{ incompatible_plugins }}"
+  when: incompatible_plugins != {}
```
mikenairn: So this will just fail and tell the person they need to update these plugins manually themselves if they want to continue?

mikenairn: Did a test and that's what is happening anyway:

```TASK [preflight_checks : fail] ***********************************************************************************************************************************************************************************************************************************************************
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "Found the following incompatible plugins, please update these to the target version. {'github': {'current_version': '1.29.3', 'target_version': '1.29.4'}}"}

PLAY RECAP *******************************************************************************************************************************************************************************************************************************************************************************
127.0.0.1                  : ok=13   changed=1    unreachable=0    failed=1    skipped=1    rescued=0    ignored=0   ```

What happens if we try and install the version of the plugin we are specifying? I'm just wondering if we should have it fail here or not.

Comment (JameelB) [reply scripts/playbooks/roles/preflight_checks/tasks/check_plugins.yaml]: @mikenairn yes. They either need to manually update these or delete them so that they get re-installed with the correct target version. Our cli for installing the plugins, skips the installation of ones that are already installed.

File: scripts/playbooks/roles/preflight_checks/tasks/check_plugins.yaml [hunk 3]
```diff
@@ -0,0 +1,45 @@
...
+  when: get_installed_plugins.stdout_lines | length > 0
+
+- pause:
+    prompt: "Incompatible plugins found. This may cause the Delorean jobs to not work as expected.\n{{ incompatible_plugins }}\nDo you want to continue? [y/n]"
+  register: user_confirmation
+  when: incompatible_plugins != {}
+
+- fail:
+    msg: "Jenkins configuration discontinued. Please update or uninstall these plugins before running this script again."
+  when: user_confirmation.user_input is defined and user_confirmation.user_input != 'y'
```
mikenairn: It now lets you continue if you want as expected :+1: 

```TASK [preflight_checks : pause] **********************************************************************************************************************************************************************************************************************************************************
[preflight_checks : pause]
Incompatible plugins found. This may cause the Delorean jobs to not work as expected.
{'ssh-agent': {'current_version': '1.17', 'target_version': '1.16'}, 'extended-choice-parameter': {'current_version': '0.78', 'target_version': '0.76'}, 'ansicolor': {'current_version': '0.6.2', 'target_version': '0.5.2'}}
Do you want to continue? [y/n]:
```

PR #216, opened 2019-06-26
INTLY-2439 add w2 test executor pipeline

tremes:

### DIFF
--- a/jobs/integr8ly/w2-test-executor.yaml
+++ b/jobs/integr8ly/w2-test-executor.yaml
@@ -1,14 +1,14 @@
 ---
 
 - job:
-    name: w1-test-executor
+    name: w2-test-executor
     project-type: pipeline
     sandbox: false
     parameters:
       - string:
           name: REPOSITORY
           default: 'https://gitlab.cee.redhat.com/integreatly-qe/integreatly-qe.git'
-          description: 'QE repository containing the smoke tests source code.'
+          description: 'QE repository containing the tests source code.'
       - string:
           name: BRANCH
           default: 'master'
@@ -52,8 +52,8 @@
                                 } // if
                             } // stage
 
-                            stage('Run create AMQ Online address test') {
-                                buildStatus = build(job: 'amq-online-address-creation-test', parameters: [
+                            stage('Run Launcher test') {
+                                buildStatus = build(job: 'launcher-tests', parameters: [
                                     string(name: 'REPOSITORY', value: "${REPOSITORY}"),
                                     string(name: 'BRANCH', value: "${BRANCH}"),
                                     string(name: 'CLUSTER_URL', value: "${CLUSTER_URL}"),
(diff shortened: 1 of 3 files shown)


tremes:
Jenkins pipeline https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/w2-test-executor/

jobs/integr8ly/w2-test-executor.yaml
```diff
@@ -0,0 +1,104 @@
...
+                                    string(name: 'REPOSITORY', value: "${REPOSITORY}"),
+                                    string(name: 'BRANCH', value: "${BRANCH}"),
+                                    string(name: 'CLUSTER_URL', value: "${CLUSTER_URL}"),
+                                    string(name: 'NAMESPACE_PREFIX', value: "${NAMESPACE_PREFIX}"),
+                                    string(name: 'ADMIN_USERNAME', value: "${ADMIN_USERNAME}"),
+                                    string(name: 'ADMIN_PASSWORD', value: "${ADMIN_PASSWORD}"),
+                                    string(name: 'EVALS_USERNAME', value: "${EVALS_USERNAME}"),
+                                    string(name: 'EVALS_PASSWORD', value: "${EVALS_PASSWORD}"),
+                                    string(name: 'TIMEOUT_THRESHOLD', value: "${TIMEOUT_THRESHOLD}"),
+                                    booleanParam(name: 'CLEAN_RESOURCES', value: true)
```
trepel: Should not this be false here? I thought it is there since the 3scale test expects some stuff to exist and this stuff is created by this Launcher test. So it should not clean the resources afterward. Or am I missing something?

tremes: We are OK with `true` in this case IMO since "clean" means only whether to delete the newly created GH repo for the booster or not. We can go with `false` but it's just very minor difference IMO. 

3scale relies "only" on running booster in an OpenShift cluster (and we don't delete the booster after test)

tremes:
> For launcher test we have standalone pipeline and for 3scale test we don't. Why this inconsistency? 

@trepel Because I am lazy ... :) and it seemed to me as a waste of time...but if you want to create it....

trepel:
@tremes OK, let's leave it as it is now

jobs/integr8ly/w2-test-executor.yaml
trepel: ah, ok, so even if the CLEAN_RESOURCES is set to true, it only removes underlying GH repository but it does not remve the booster itself. Ack, in that case 'true' can indeed be here.

tremes: yes that's the case. We can extend this (to also delete the booster in OS) in the future.

tremes:
Ohh yeah I always forget....

PR #217, opened 2019-06-27
Add new w2-executor into the all-tests-executor definition

tremes:


tremes:
Follow up for INTLY-2439

PR #223, opened 2019-07-04
add task to configure delorean jobs

JameelB:
## What
- [x] Create a role for configuring delorean jobs
   - This should create a jenkins-config.ini file with Jenkins configuration
   - This should call the `scripts/configure_jenkins.sh` script to create the delorean jobs in the target Jenkins instance.
- [x] Include this role in the `configure_jenkins.yaml` playbook
- [x] Create an additional playbook to configure the delorean jobs only

The playbook is configured to skip the creation of delorean  jobs by default. In order to create the delorean jobs, the variable `configure_delorean_jobs` must be set to true.

Example command: 
```
ansible-playbook -i inventories/hosts playbooks/configure_jenkins -e configure_delorean_jobs=true
```

### DIFF
--- a/scripts/playbooks/configure_jenkins.yaml
+++ b/scripts/playbooks/configure_jenkins.yaml
@@ -14,3 +14,7 @@
     - name: Configure Global Settings
       include_role:
         name: configure_global_settings
+    - name: Configure Delorean Jobs
+      include_role:
+        name: configure_delorean_jobs
+      when: configure_delorean_jobs is defined and configure_delorean_jobs == true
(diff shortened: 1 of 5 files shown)


scripts/playbooks/configure_jenkins.yaml
```diff
@@ -13,4 +13,8 @@
         name: add_credentials
     - name: Configure Global Settings
       include_role:
-        name: configure_global_settings
\ No newline at end of file
+        name: configure_global_settings
+    - name: Configure Delorean Jobs
+      include_role:
+        name: configure_delorean_jobs
+      when: configure_delorean_jobs is defined and configure_delorean_jobs == true
```
mikenairn: Default is to not configure jobs :+1:

PR #227, opened 2019-07-08
Misc updates

mikenairn:
## What
50fbde24212b97bb0128d1ce6b0fdd29dd4dcfcb Add ws-cleanup:0.37 plugin
8535bfade1130c03c8e3acfb018cbba726be842c Add deploy_jenkins script to provision a jenkins instance on an openshift cluster.
a5cd1b6eb95168f079d86f09d12943715b84b32e Update group_vars files to match how we have them setup in the credentials repo.
e5c0c40d27519f15009c908f037561355939a214 Add jenkins s2i configuration to scripts/s2i Moved plugins.txt to scripts/s2i and updated tasks and docs accordingly.
5916f045b1b9f3422015ee4501bcf5a584642f5f Fix credentials tasks if csrf is false.

Verification:
* Follow docs here 2ca42be49ae4a701d3dd89a4a51466398ced815c


docs/delorean/deploy-jenkins.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Table of Contents
+  - [1. Prerequisites](#prerequisites)
+  - [2. Deploy Jenkins on OpenShift](#deploy-on-openshift)
+
+
+## Prerequisites
+- OpenShift 3.x Cluster (oc login)
+
+## Deploy on OpenShift
```
JameelB: Can make this the same as the heading in the table of contents? `Deploy Jenkins on OpenShift`

scripts/deploy_jenkins.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh -xe
+SCRIPTS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+TEMPLATES_DIR="$( cd $SCRIPTS_DIR/templates && pwd )"
+
+oc new-project $PROJECT_NAME
+
+for AGENT in jenkins-tools
```
JameelB: Where is this `jenkins-tools` coming from?

docs/delorean/deploy-jenkins.md
mikenairn: Matches now

scripts/deploy_jenkins.sh
mikenairn: This is just an array with 1 thing in "jenkins-tools", if you had more than 1 agent you could add it here. jenkins-tools is a container we used previously that has basic tools pre installed on it, one of which is skopeo which is all we really rely on for the openshift jobs https://github.com/feedhenry/wendy-jenkins-s2i-configuration/blob/master/agent-jenkins-tools/Dockerfile

The container is already built and available on the feedhenry quay repo, i never bothered changing it, so it's just referenced here in the template https://github.com/integr8ly/ci-cd/blob/13ad5dad09eb569719428d4ddfbc539c67956835/scripts/templates/agent-image-template.yml

When Jenkins starts up, the kuberenetes plugin will check for ImageStreams available that have the correct role label applied https://github.com/integr8ly/ci-cd/blob/13ad5dad09eb569719428d4ddfbc539c67956835/scripts/templates/agent-image-template.yml#L22 
If it finds one, it will add that as k8s pod template in the global config  

image

If you request a node matching those labels it will spin up a new pod in OpenShift and run everything inside of that.

mikenairn: If it helps, this is where it was taken from https://github.com/feedhenry/wendy-jenkins-s2i-configuration/blob/master/scripts/deploy_jenkins.sh#L50

PR #228, opened 2019-07-09
added 3scale restoration pipeline

pawelpaszki:
## What
Create a pipeline definition for 3scale-restoration test

### DIFF
--- a/jobs/integr8ly/three-scale-restoration.yaml
+++ b/jobs/integr8ly/three-scale-restoration.yaml
@@ -1,0 +1,136 @@
+---
+
+- job:
+    name: three-scale-restoration
+    project-type: pipeline
+    sandbox: false
+    parameters:
+      - string:
+          name: REPOSITORY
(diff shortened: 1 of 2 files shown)


jobs/integr8ly/three-scale-restoration.yaml
```diff
@@ -0,0 +1,136 @@
...
+          name: REPOSITORY
+          default: 'https://gitlab.cee.redhat.com/integreatly-qe/integreatly-qe.git'
+          description: 'QE repository containing the tests source code.'
+      - string:
+          name: BRANCH
+          default: 'master'
+          description: 'Branch of the repository'
+      - string:
+          name: CLUSTER_URL
+          description: 'URL of cluster on which the W1 test will be executed.'
```
trepel: Wrong description. Replace with e..g `URL of the cluster with Integreatly backups installed`

pawelpaszki: good catch. Thanks!

PR #233, opened 2019-07-11
Update RC job

mikenairn:
Update RC job to check for images in a project tech-preview repo if it fails to find it in the project repo.

This was required for fuse where one of the images wasn't in the expected repo in brew.

### DIFF
--- a/jobs/openshift/cluster/brew/Jenkinsfile
+++ b/jobs/openshift/cluster/brew/Jenkinsfile
@@ -4,14 +4,14 @@
 def imageSyncOptions = [:]
 
 def verifyImageSyncOptions(options) {
-    if(!options.openshiftMasterUrl || !options.clusterAdminUsername || !options.clusterAdminPassword || !options.installationGitBranch || !options.installationGitUrl) {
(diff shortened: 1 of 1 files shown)


JameelB:
Shouldn't the `integreatly.yml` be configured to catch these? So in cases of an image being in a different repo, it should have an additional project like so: 

```
imageSync:
  sourceRepo: brew-pulp-docker01.web.prod.ext.phx2.redhat.com:8888
  projects:
    fuse7:
      images:
      - fuse-ignite-meta:1.4-1
      - fuse-ignite-s2i:1.4-1
      - fuse-ignite-server:1.4-2
      - fuse-ignite-ui:1.4-1
      - fuse-online-operator:1.4-3
    fuse7-tech-preview:
      - fuse-postgres-exporter:1.4-1
```

mikenairn:
@JameelB Yes it should, but in the time i had this was the best i could come up with. The problem is there's a mismatch between the project in brew (where the image actually is "fuse7-tech-preview") and the template (where is expects to find it "fuse7"). If i change it to be like what you are suggesting, the image sync part will work, but it will fail to install as the template will be looking for the image in "fuse7", but we pushed it to "fuse7-tech-preview"

PR #238, opened 2019-07-19
Add amq-broker-7 to amq online rc registry projects to search

mikenairn:


mikenairn:
@david-martin FYI, this will need to be merged so that the next time the rc discovery job runs it includes this registry in it's search

PR #242, opened 2019-07-25
added install backup ns feature to pds install related jobs and triggers

pawelpaszki:
## What
added install backup ns feature to pds install related jobs and triggers

## Verification
1. Go to this pipeline: https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/backups-pds-install/25/console and confirm that at the beginning of the build the backups namespace and secret get created (search for: **namespace/openshift-integreatly-backups created** and **secret/s3-credentials created**
2. After the pipeline completes, login to the cluster, on which integreatly was installed and confirm that the backups namespace is present.
3. Run any of one time jobs and make sure that the job completes, e.g. s3 url for the backed up files is available

Alternatively run the job yourself and go through verification step 2 and 3

### DIFF
--- a/jobs/integr8ly/pds-install.yaml
+++ b/jobs/integr8ly/pds-install.yaml
@@ -38,6 +38,7 @@
             description: "User capable of SSH-ing to bastion server"
         - string:
             name: BASTION_PRIVATE_KEY_ID
+            default: pds-bastion-pem
             description: "ID of SSH Credentials (private key) used for SSH-ing to bastion"
         - string:
             name: ADDITIONAL_ANSIBLE_PARAMS
@@ -46,6 +47,10 @@
             name: PATCH_TO_MASTER
             default: false
             description: "Indicates whether the components should be patched to master during installation"
+        - bool:
+            name: INSTALL_BACKUPS
+            default: false
+            description: "Indicates whether backups namespace should be installed during Integr8ly installation"
         - string:
             name: NAMESPACE_PREFIX
             description: "This value will be prefixed to the names of the namespaces created during Integr8ly installation. Defaulting to empty string for RHPDS"
@@ -66,10 +71,11 @@
                                   "backup_version"]
         String MASTER_URL = "master1.${YOURCITY}.internal"
         String BASTION_URL = "bastion.${YOURCITY}.openshiftworkshop.com"    
-        String bastionLabel = "${BASTION_URL}-slave"                      
-        
+        String bastionLabel = "${BASTION_URL}-slave"   
+        String secretFileName = "QE-AWS-SECRET.yml"                   
+        def secretContent
         try {
-            timeout(60) { ansiColor('gnome-terminal') { timestamps {
+            timeout(75) { ansiColor('gnome-terminal') { timestamps {
                 node('cirhos_rhel7') {       
                     stage('Verify input') {
                         if (!YOURCITY) {
@@ -115,6 +121,12 @@
                             println "Slave ${bastionLabel} has been created successfully."
                         } // end if
                     } // stage
+
+                    stage('Obtain s3-credentials secret for backups namespace') {
+                        if (INSTALL_BACKUPS.toString() == 'true') {
+                            secretContent = sh (returnStdout: true, script: 'curl -k https://gitlab.cee.redhat.com/integreatly-qe/integreatly-qe/raw/development/resources/QE-AWS-SECRET.yml')
+                        }
+                    } // stage
                 } // node
 
                 node(bastionLabel) {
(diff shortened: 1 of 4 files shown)


jobs/integr8ly/pds-install.yaml
```diff
@@ -162,6 +188,27 @@
...
+                                
+                                def secrets = sh(script: obtainSecretsCommand, returnStdout: true)
+                                if (!secrets.contains(secretName)) {
+                                    writeYaml file: secretFileName, data: secretContent
+                                    sh """
+                                        sudo oc create -f ${secretFileName}
+                                        rm ${secretFileName}
+                                    """
+                                }
+                            }
```
tremes: Wouldn't be easier just to try to create the secret (with ` sudo oc create -f ${secretFileName}`) and optionally catch error and log some message in case of already existing secret?

jobs/integr8ly/pds-install.yaml
```diff
@@ -162,6 +188,27 @@
                                     """
                                 }
                             }
+                            if (INSTALL_BACKUPS.toString() == 'true') {
+                                String backupsNamespace = 'openshift-integreatly-backups';
```
tremes: Is this always with `openshift-` prefix ?

jobs/integr8ly/pds-install.yaml
pawelpaszki: Probably would be easier, not sure if cleaner though. I don't have any strong preference

jobs/integr8ly/pds-install.yaml
pawelpaszki: yes

trepel: yep, I think we can have this hard-coded. There is a param for install playbook to change this value, however, effectively only this default value is used.

jobs/integr8ly/pds-install.yaml
trepel: If I got it right the TEST_SUITE_REPOSITORY and TEST_SUITE_BRANCH are only used for secret creation. It seems that `curl -k <rawGitLabUrlToSecretFile>` is sufficient to get the `secretContent` (see https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/trepel-pipeline-playground/42/console). You just need to run it from `cirhos_rhel7`, not bastion. So you don't need those params at all.

pawelpaszki: It is probably ok on public GH repo, however on GItLab you need to provide your token to get file this way

pawelpaszki:
@trepel can we merge it?

jobs/integr8ly/pds-install.yaml
trepel: Not sure if I understand. I think this is the simplest way to get the `secetContent`:
```
String secretContent = sh (returnStdout: true, script: 'curl -k https://gitlab.cee.redhat.com/integreatly-qe/integreatly-qe/raw/development/resources/QE-AWS-SECRET.yml')
println secretContent
```

No need to clone the GitLab repo. But this piece of code needs to be (not sure) executed from `cirhos_rhel7`, not sure if that would work from bastion.

pawelpaszki: I think you are right. Will update it

pawelpaszki: @trepel mind having another look?

jobs/integr8ly/pds-install.yaml
```diff
@@ -139,6 +151,7 @@
                     stage('Prepare environment') {
                         dir('installation') {
                             sh """
+                                sudo oc login -u system:admin
```
trepel: why this command is here? You only play with the files so no need to be logged in?

And this runs on bastion directly so I think you are logged in as system:admin by default, aren't you?

pawelpaszki: this command was moved from below to be able to create a secret and ns (I believe it is still needed here)

jobs/integr8ly/pds-testing-nightly-trigger.yaml
```diff
@@ -29,6 +29,7 @@
             CUSTOMER_ADMIN_PASSWORD=Password1
             CLEAN_RESOURCES=true
             MANIFEST_VERSION=master
+            INSTALL_BACKUPS=true
```
trepel: Do we want to have this set to `true` by default?

pawelpaszki: no preference. having it set to true will ease backup and restore testing. on the other hand might increase s3 usage

jobs/integr8ly/pds-install.yaml
trepel: hmm, strange, but ok :)

jobs/integr8ly/pds-testing-nightly-trigger.yaml
trepel: I would go with `false` by default to be honest

PR #247, opened 2019-08-07
use password input for cluster admin password

JameelB:
## What
Use password input for the cluster admin password parameter instead of plaintext


JameelB:
ping @mikenairn @pmccarthy

PR #249, opened 2019-08-15
INTLY-2841 mark browser tests as unstable if there's some test failure

tremes:


tremes:
Runs are https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/browser-based-single-test/125/
and 
https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/browser-based-testsuite-pipeline/828/

PR #252, opened 2019-08-20
PDS pipelines - parameterize the cluster domain

tremes:


trepel:
@tremes What about having that `open.redhat.com` as default? I think that this `example.opentlc.com` is used only as an interm solution for let's encrypt issue.

tremes:
@trepel yes could be

trepel:
@tremes Should not be CLUSTER_DOMAIN be added when you call pds-install from within the pds-testing-executor?

tremes:
@trepel yep. Thanks

tremes:
Running here https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/pds-testing-executor/409/

PR #256, opened 2019-08-23
INTLY-2944 fix the conditions to set correct build status

tremes:


tremes:
I hope this should be sufficient to fix the problem...

tremes:
OK so this doesn't help ... :-/

PR #262, opened 2019-09-09
INTLY-2816 - split up the browser tests execution

tremes:

### DIFF
--- a/jobs/integr8ly/ocp3/test-suites/all-tests-executor.yaml
+++ b/jobs/integr8ly/ocp3/test-suites/all-tests-executor.yaml
@@ -94,6 +94,7 @@
             string(name: 'NUMBER_OF_USERS', value: "${NUMBER_OF_USERS}"),
         ]
         def err = null
+        def wtTestStatus;
         try {
             timeout(180) { ansiColor('gnome-terminal') { timestamps {
                 node('cirhos_rhel7') {
@@ -107,7 +108,7 @@
                     }
 
                     stage ('Start Basic Walkthroughs Browser Test') {
-                        runSingleBrowserTest('tests/00_walkthroughs_start/start-walkthroughs.js')
+                       wtTestStatus = runSingleBrowserTest('tests/00_walkthroughs_start/start-walkthroughs.js')
                     } // stage
 
                     stage ('After First Login Tests') {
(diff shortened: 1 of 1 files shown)


tremes:
I don't say this is the best solution but I think it's more granular and more clear (once there is some failure)

tremes:
Executed here https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/tremes-all-tests-executor/1/console for PoC testing.

jobs/integr8ly/ocp3/test-suites/all-tests-executor.yaml
```diff
@@ -152,3 +167,13 @@
                 currentBuild.result = 'UNSTABLE'
             }
         }
+
+           def runSingleBrowserTest(testName) {
```
pawelpaszki: Please fix the inconsistent spacing

pawelpaszki:
would it not be better to get the result of 00_walkthroughs test and don't execute wt-related tests in case of failure?

jobs/integr8ly/ocp3/test-suites/all-tests-executor.yaml
tremes: oh yeah..thanks

tremes:
@pawelpaszki interesting idea...maybe yes. Let's see what others think.

damienomurchu:
I like the changes, and agree with suggestion to not run a wt-related test if the WT is not in the solution explorer. If one walkthrough is unavailable, but the others are, we should still test the run the WT test(s) for the WT(s) that are there though. 

That probably means splitting out `00_walkthroughs_start` as that currently tests for presence of all WT's.

trepel:
To be honest I would leave this as is and don't do any skipping of other browser tests just because `00_walkthroughs` test failed.

If we want to make our life easier we can make sure the test fails early if some prereqs are not met. Ideally with a message like this one: `This test expects this and this to be in place but it seems it is not the case.` But this holds for each and every test, not just the browser ones.

trepel:
We can skip the wt test if wt is not present - no issues with that. But in that case, we need to have separate test checking the presence of each walkthrough. So that there is some failed test in case some wt is missing (not just skipped test, that is easier to overlook).

tremes:
I added one more commit. Did you mean something like that @pawelpaszki pls? And what others think please?

trepel:
I think this one is good to go :)

tremes:
In theory we could also skip the w1 and w2 executors if there's some failure in the start walkthroughs test. WDYT?

pawelpaszki:
that was my original thought - to skip all wt1 and wt2 related (backend and ui) tests

trepel:
Is there any (sub)test in wt1 or wt2 executor that would pass even if start walkthroughs test failed? I don't think so, but you might know. If yes, there would still be some value in running them.

Also, there is an option that start walkthroughs test fails at the very end but it does effectively start them - not highly probable but still.

I would slightly prefer to fail early saying the prereqs were not met instead of skipping.

tremes:
The w2 and w1 executor tests require existence of related "shared services" namespace and also w1 or  w2 namespace. So there won't pass any test when these don't exist.

Second option can probably happen, but it's unlikely IMO (maybe some timeout could be likely). 

I don't have a strong opinion here. I can update it or leave as it is....as you want guys.

trepel:
no strong opinion either. I guess it is you who is to do the actual work so up to you really :)

tremes:
Let's be optimistic and believe that the start walkthrough test won't fail so often :) I am leaving it in current shape.

PR #271, opened 2019-09-24
Minor improvement + sleep 2 min in all-tests-executor after starting …

tremes:
…walkthroughs

## What
- Display kubeadmin password for each OS4 cluster create run
- Sleep 2 min before running after-first-login


tremes:
@trepel I think the password is generated and available only in the pipeline output. Correct if me I am wrong...

PR #273, opened 2019-09-25
moved dedicated-admin install to ocp4-install and allowed using custom password

pawelpaszki:
## What
moved dedicated-admin install to ocp4-install and allowed using custom password

## Verification
The pipeline with changed definition was executed here: https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/ppaszki-ocp4-install/45/console


tonyxrmdavidson:
Checked that customer-admin can be used to login to cluster with oc login. Successful.

PR #274, opened 2019-09-25
Added CLEAN_RESOURES to wt1 executor

trepel:
## What

Added CLEAN_RESOURCES param to wt1 executor (similarly to other tests). Useful if you want to keep the resources for subsequent manual work or (for instance) upgrade.

This should also make sure that if you set CLEAN_RESOURCES to 'false' in all-test-executor, it is properly propagated to wt1 executor.

Successful Jenkins run:
https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/Test%20Suites/job/w1-test-executor/532/ (config from this PR has been used).

Verification Steps:
probably just eye review, we have these pieces all over the place.


tremes:
I noticed this problem too.

PR #278, opened 2019-09-27
SSO test - parameterize user name

tremes:


tremes:
Test run https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/sso-user-create-tests/535/.

I don't think we need to update pds-testing-executor and poc-testing-executor. I think we can just rely on default value there.

pawelpaszki:
Looks ok (although using param like `USER_PREFIX` might have been more suitable, but don't mind really)

trepel:
I can't really understand how you learn about incoming PR so quickly @pawelpaszki :)

Yes, I would have exactly the same suggestion, `USERNAME_PREFIX` is even better, but minor and not a blocker for merging.

tremes:
Prefix? Come on guys what does it prefix :) ? The poor number https://gitlab.cee.redhat.com/integreatly-qe/integreatly-qe/blob/master/test-suites/backend-testsuite/tests/sso/create-users.js#L31 ? :)

tremes:
But if you like it more I can change it :)

trepel:
Haha, I like it more, it is more exact than just "USERNAME". Look, you can leave it as is, but at least improve the description then.

`Name of newly created users in SSO, 01 ... XX numbers will be appended to assemble the usernames.`

PR #287, opened 2019-10-02
changed default value of ocp4-testing-executor cleanup flags

pawelpaszki:
## What
changed default value of ocp4-testing-executor cleanup flags


tonyxrmdavidson:
This looks fine.  LGTM.

PR #290, opened 2019-10-03
INTLY-3680 introduce OSD install pipeline

tremes:

### DIFF
--- a/jobs/openshift/cluster/integreatly/osd-cluster-integreatly-install.yaml
+++ b/jobs/openshift/cluster/integreatly/osd-cluster-integreatly-install.yaml
@@ -1,0 +1,95 @@
+---
+- job:
+    name: osd-cluster-integreatly-install
+    display-name: 'OSD Cluster Integreatly install'
+    project-type: pipeline
+    concurrent: false
+    parameters:
+      - string:
+          name: 'clusterName'
+          default: ''
+          description: '[REQUIRED] The name of the target cluster to install Integreatly against'
+      - string:
+          name: 'installationGitBranch'
+          default: 'master'
+          description: '[REQUIRED] The name of the git branch to be used for the Integreatly installer'
+      - string:
+          name: 'routerShard'
+          default: ''
+          description: '[REQUIRED] Router shard of the targeted cluster'
+      - string:
+          name: 'openShiftToken'
+          default: ''
+          description: '[REQUIRED] Openshift token for cluster-admin user'
+      - string:
+          name: 'githubClientId'
+          default: ''
+          description: '[REQUIRED] Github client id for launcher authentication'
+      - string:
+          name: 'githubClientSecret'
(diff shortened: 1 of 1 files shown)


tremes:
The pipeline is available at https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/OSD/job/osd-cluster-integreatly-install. I haven't had a chance to run it yet.

jobs/openshift/cluster/integreatly/osd-cluster-integreatly-install.yaml
```diff
@@ -0,0 +1,95 @@
...
+                  jobTemplate: 'Integreatly_Bootstrap_and_Install_[OSD]',
+                  templateType: 'workflow',
+                  importTowerLogs: true,
+                  removeColor: false,
+                  verbose: true,
+                  extraVars: """---
+                    tower_environment: dev
+                    rhsso_namespace: openshift-sso
+                    cluster_name: ${clusterName}
+                    integreatly_osd_uninstall_branch: ${uninstallationGitBranch}
```
pawelpaszki: should it be `installationGitBranch` (both extra var and the value)?

tremes: ahh good catch. Thanks

jobs/openshift/cluster/integreatly/osd-cluster-integreatly-install.yaml
```diff
@@ -0,0 +1,95 @@
...
+                    router_shard: ${routerShard}
+                    openshift_token: ${openShiftToken}
+                    github_client_id: ${githubClientId}
+                    github_client_secret: ${githubClientSecret}
+                    rhsso_admin_pass: ${customerAdminPassword}
+                    threescale_s3_access_key: ${3scaleAwsAccessKey}
+                    threescale_s3_secret_key: ${3scaleAwsAccessKey}
```
pawelpaszki: I think the value here should be: `3scaleAwsSecretKey`

pawelpaszki:
see my comments. I will approve after successful execution (or at least failure not caused by the pipeline)

jobs/openshift/cluster/integreatly/osd-cluster-integreatly-install.yaml
tremes: Ohh yeah week is approaching its end... :) Thanks.

tremes:
Pipeline parameters can't start with a number. Updated.

trepel:
LGTM, but it would be good to see successful execution. Try to reach out to @psturc , he might be in need to install on OSD so this job might be used instead of doing it directly via tower.

trepel:
@psturc cool, thank you. It seems this is good to go then. Merging.

PR #297, opened 2019-10-09
OSD install pipeline - update client secret in RHSSO automatically

tremes:


tremes:
I don't have a successful run yet (because I screwed up my first attempt)

jobs/openshift/cluster/integreatly/osd-cluster-integreatly-install.yaml
```diff
@@ -92,4 +95,19 @@
...
+            dir('.') {
+              git branch: "master", url: "https://gitlab.cee.redhat.com/integreatly-qe/integreatly-qe.git"
+            }
+          }
+
+          stage('Update client secret in SSO') {
+            dir('test-suites/backend-testsuite') {
+              sh '''
+                npm install
+                node ssoClientUpdate.js ${clientSecret}
```
pawelpaszki: I think there are some parameters required to run this script (SSO_URL, CLUSTER_URL and OPENSHIFT_TOKEN). I can't see them anywhere here. Am I missing something?

tremes: Oh yes you're right. I totally forgot about them. The token is available...but not passed. And others (especially the SSO_URL) are missing. Thanks

tremes: Well we'll see if it works...

jobs/openshift/cluster/integreatly/osd-cluster-integreatly-install.yaml
```diff
@@ -92,4 +95,22 @@
                 )
             }  
           }
+          
+          stage('Clone QE repository') {
+            dir('.') {
+              git branch: "master", url: "https://gitlab.cee.redhat.com/integreatly-qe/integreatly-qe.git"
```
tremes: Do we want to parameterize that? I don't think it's necessary here....

pawelpaszki: hardcoded is fine (if you are referring to the integreatly-qe branch and repo)

tremes:
Finally a successful run https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/OSD/job/osd-cluster-integreatly-install-tremes/3/

trepel:
The values might be in the param description but I am not sure if that's the best solution, any thoughts?

PR #300, opened 2019-10-10
changed pipelines triggering times

pawelpaszki:
## What
changed pipelines triggering times (and type of installation) due to automatic PDS instances stopping and 3scale test timezone related problems


tremes:
AFAIK only the `delorean-testing-nightly-trigger` is being recreated automatically so it's better to update the pds triggers directly in Jenkins AFAIK.

pawelpaszki:
I have already updated all of them. Just wanted to keep them in sync

tremes:
OK cool. Thx

PR #307, opened 2019-10-11
INTLY-2401 introduce first version of nightly OSD tesing pipeline

tremes:

### DIFF
--- a/jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
+++ b/jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
@@ -1,0 +1,217 @@
+---
+
+- job:
+    name: osd-testing-nightly-trigger
+    project-type: pipeline
+    description: "Nightly pipeline trigger for QE OSD cluster (rhmi-qe1) testing."
+    sandbox: false
+    concurrent: false
+    triggers:
+      - timed: "H(20) 3 * * *"
+    properties:
+      - build-discarder:
+          num-to-keep: 56
+      - inject:
+          properties-content: |
+            CLUSTER_NAME=rhmi-qe1
+            RECIPIENTS=integreatly-qe@redhat.com
+            INSTALLATION_BRANCH=master
+            ROUTER_SHARD=f2d1
+            USERNAME=osdtester
+            TEST_SUITES_REPOSITORY=https://gitlab.cee.redhat.com/integreatly-qe/integreatly-qe.git
+            TEST_SUITES_BRANCH=master
+            GH_CLIENT_ID=836affcd7fd1bcebf9ae
+            GH_CLIENT_SECRET=72e97db6f1a2a6d8d2be66686cf52ccf15b3588f
+            CUSTOMER_ADMIN_PASSWORD=Password1
+            CREDENTIALS_ID=osd-s3-access-key-and-secret
+            THREESCALE_BUCKET_NAME=integreatly-qe1-3scale
+            THREESCALE_BUCKET_REGION=eu-west-1
+            BACKUPS_BUCKET_NAME=rhmi-qe1-backups
+            ADMIN_USERNAME=admin01
+            ADMIN_PASSWORD=Password1
+            CLEAN_RESOURCES=false
+            MANIFEST_VERSION=master
+            TESTING_MASTER=false
+            TIMEOUT_THRESHOLD=1
+    dsl: |
+        def err = null
+        def loggedIn = false
+        def installationStatus 
+        def uninstallationStatus
+        try {
+            timeout(300) { ansiColor('gnome-terminal') { timestamps {
+                node('cirhos_rhel7') {
+                    currentBuild.displayName = "${currentBuild.displayName} ${CLUSTER_NAME}"
+                    currentBuild.description = "branch: ${INSTALLATION_BRANCH}"
+
+                    stage ('Run uninstallation pipeline') {
+                        withCredentials([
+                                [$class: 'StringBinding', credentialsId: "rhmi-qe1-token", variable: 'OPENSHIFT_TOKEN']
+                            ]) {
+                            uninstallationStatus = build(job: 'osd-cluster-integreatly-uninstall', propagate: false, parameters: [
+                                string(name: 'clusterName', value: "${CLUSTER_NAME}"),
+                                string(name: 'uninstallationGitBranch', value: "${INSTALLATION_BRANCH}"),
+                                string(name: 'routerShard', value: "${ROUTER_SHARD}"),
+                                string(name: 'openShiftToken', value: "${OPENSHIFT_TOKEN}"),
+                                string(name: 'userName', value: "${USERNAME}")
+                            ]).result
+                        }
+                    } // stage
+
+                     if (uninstallationStatus != 'SUCCESS') {
+                        println 'There were some errors in previous uninstallation pipeline!'
+                        currentBuild.result = 'FAILURE'
+                        return
+                    }
+
+                    stage ('Run installation pipeline') {
+                        withCredentials([
+                                [$class: 'AmazonWebServicesCredentialsBinding', credentialsId: "${CREDENTIALS_ID}", accessKeyVariable: 'AWS_ACCESS_KEY_ID', secretKeyVariable: 'AWS_SECRET_ACCESS_KEY'],
+                                [$class: 'StringBinding', credentialsId: "rhmi-qe1-sso-client-secret", variable: 'SSO_SECRET'],
+                                [$class: 'StringBinding', credentialsId: "rhmi-qe1-token", variable: 'OPENSHIFT_TOKEN']
+                            ]) {
+                            installationStatus = build(job: 'osd-cluster-integreatly-install', propagate: false, parameters: [
+                                string(name: 'clusterName', value: "${CLUSTER_NAME}"),
+                                string(name: 'installationGitBranch', value: "${INSTALLATION_BRANCH}"),
+                                string(name: 'routerShard', value: "${ROUTER_SHARD}"),
+                                string(name: 'openShiftToken', value: "${OPENSHIFT_TOKEN}"),
+                                string(name: 'githubClientId', value: "${GH_CLIENT_ID}"),
+                                string(name: 'githubClientSecret', value: "${GH_CLIENT_SECRET}"),
+                                string(name: 'customerAdminPassword', value: "${CUSTOMER_ADMIN_PASSWORD}"),
+                                string(name: 'threeScaleAwsAccessKey', value: "${AWS_ACCESS_KEY_ID}"),
+                                string(name: 'threeScaleAwsSecretKey', value: "${AWS_SECRET_ACCESS_KEY}"),
+                                string(name: 'threeScaleBucketName', value: "${THREESCALE_BUCKET_NAME}"),
+                                string(name: 'threeScaleBucketRegion', value: "${THREESCALE_BUCKET_REGION}"),
+                                string(name: 'backupAwsKey', value: "${AWS_ACCESS_KEY_ID}"),
+                                string(name: 'backupAwsSecret', value: "${AWS_SECRET_ACCESS_KEY}"),
+                                string(name: 'backupsBucketName', value: "${BACKUPS_BUCKET_NAME}"),
+                                string(name: 'clientSecret', value: "${SSO_SECRET}")
+                            ]).result
+                        }
+                    } // stage
+
+                    if (installationStatus != 'SUCCESS') {
+                        println 'There were some errors in previous installation pipeline!'
+                        currentBuild.result = 'FAILURE'
+                        return
+                    }
+
+                    // This relies on a cronjob to restart the OpenShift API so that we can successfully login afterwards
+                    stage ('Try to login to installed cluster') {
+                        println "Waiting for successful login to the ${CLUSTER_NAME} cluster"
+                        i = 0
+                        // Trying to log into the cluster 12 times and each 5 minutes which can take max 1 hour 
+                        while(!loggedIn && i < 12) {
+                            loggedIn = ocLoginAsCustomerAdmin()
+                            i += 1
+                            sleep time: 5, unit: 'MINUTES'
+                        }
+                    }
+
+                    // we want to run test only when there was a successful login to the cluster
+                    if (loggedIn) {
+
+                        stage ('Create admin user in SSO') {
+                            build(job: 'sso-user-create-tests', propagate: false, parameters: [
+                                string(name: 'REPOSITORY', value: "${TEST_SUITES_REPOSITORY}"),
+                                string(name: 'BRANCH', value: "${TEST_SUITES_BRANCH}"),
+                                string(name: 'CUSTOMER_ADMIN_USERNAME', value: "customer-admin"),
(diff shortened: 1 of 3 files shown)


jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
```diff
@@ -0,0 +1,147 @@
...
+                            string(name: 'CUSTOMER_ADMIN_USERNAME', value: "customer-admin"),
+                            string(name: 'CUSTOMER_ADMIN_PASSWORD', value: "${CUSTOMER_ADMIN_PASSWORD}"),
+                            string(name: 'SSO_URL', value: "${SSO_URL}"),
+                            string(name: 'USERNAME', value: "${USERNAME}"),
+                            string(name: 'NUMBER_OF_USERS', value: "1"),
+                            booleanParam(name: 'CLEAN_RESOURCES', value:  Boolean.valueOf("false")),
+                        ])
+                    } // stage
+
+                /**    stage ('Run tests') {
```
tremes: Test stage is temporarily excluded, because we don't have a working admin account in `oc`. We need either a permanent admin account in `oc` (there's a ticket for that and this is probably the easier solution) or to re-create an admin in SSO and delete it in `oc` and login via SSO to cluster to recreate the user in `oc`.

jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
```diff
@@ -0,0 +1,147 @@
...
+    triggers:
+      - timed: "H(0-10) 6 * * *"
+    properties:
+      - build-discarder:
+          num-to-keep: 56
+      - inject:
+          properties-content: |
+            CLUSTER_NAME=rhmi-qe1
+            RECIPIENTS=integreatly-qe@redhat.com
+            INSTALLATION_REPOSITORY=https://github.com/integr8ly/installation.git
```
pawelpaszki: I can't see any usage of this param

jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
```diff
@@ -0,0 +1,147 @@
...
+            INSTALLATION_BRANCH=master
+            ROUTER_SHARD=f2d1
+            OPENSHIFT_TOKEN=
```
pawelpaszki: is there any point in providing empty value here?

jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
```diff
@@ -0,0 +1,147 @@
...
+                            string(name: 'REPOSITORY', value: "${TEST_SUITES_REPOSITORY}"),
+                            string(name: 'BRANCH', value: "${TEST_SUITES_BRANCH}"),
+                            string(name: 'CLUSTER_URL', value: "https://console.${CLUSTER_NAME}.openshift.com"),
+                            string(name: 'EVALS_USERNAME', value: "${USERNAME}01"),
+                            string(name: 'ADMIN_USERNAME', value: "${ADMIN_USERNAME}"),
+                            string(name: 'ADMIN_PASSWORD', value: "${ADMIN_PASSWORD}"),
+                            string(name: 'CUSTOMER_ADMIN_USERNAME', value: "customer-admin"),
+                            string(name: 'CUSTOMER_ADMIN_PASSWORD', value: "${CUSTOMER_ADMIN_PASSWORD}"),
+                            string(name: 'NAMESPACE_PREFIX', value: "openshift-"),
+                            string(name: 'WEBAPP_URL', value: "${WEBAPP_URL}"),
```
pawelpaszki: can webapp url be constructed the same way as cluster url? (e.g. without providing the hardcoded value)?

jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
tremes: Yep good catch. This can be removed.

jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
```diff
@@ -0,0 +1,147 @@
...
+                    stage ('Create test user') {
+                        build(job: 'sso-user-create-tests', propagate: false, parameters: [
+                            string(name: 'REPOSITORY', value: "${TEST_SUITES_REPOSITORY}"),
+                            string(name: 'BRANCH', value: "${TEST_SUITES_BRANCH}"),
+                            string(name: 'CUSTOMER_ADMIN_USERNAME', value: "customer-admin"),
+                            string(name: 'CUSTOMER_ADMIN_PASSWORD', value: "${CUSTOMER_ADMIN_PASSWORD}"),
+                            string(name: 'SSO_URL', value: "${SSO_URL}"),
+                            string(name: 'USERNAME', value: "${USERNAME}"),
+                            string(name: 'NUMBER_OF_USERS', value: "1"),
+                            booleanParam(name: 'CLEAN_RESOURCES', value:  Boolean.valueOf("false")),
```
pawelpaszki: the comma at the end is redundant

jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
tremes: Well yes. We should probably use single style - either hardcode everything in params section or construct it as e.g `CLUSTER_URL`

pawelpaszki:
Other than minor comments it looks ok. Any chance of seeing it running?

tremes:
@pawelpaszki Thanks for quick review. Sure. It's at https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/OSD/job/osd-testing-nightly-trigger/. All the tests (they hadn't been excluded) failed in the last two runs because of the authentication issue (related to my comment about admin account).

jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
trepel: Yes, for the time being this will the same as with PDS, we need to update values in PDS nightly triggers too.

jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
trepel: yep +1 for unification. One small advantage of CLUSTER_NAME is that it is much nicer in currentBuild.description than full URL.

trepel: Not sure if we want to add that currentBuild.description to this nightly trigger. Probably nice to have.

jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
trepel: why `/**` instead of just `/*` ? Probably no harm though.

tremes: Old Java habit I guess.... :)

tremes:
I think this should be hopefully ready to go. This will need one more run to prove it's OK. Can you guys pls take a look again?

tremes:
Running here https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/Nightly/job/osd-testing-nightly-trigger/21/

jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
```diff
@@ -0,0 +1,206 @@
...
+                        currentBuild.result = 'FAILURE'
+                        return
+                    }
+
+                    // This relies on a cronjob (or something else) to update the client secret in master-config.yaml so that we can successfully login afterwards
+                    stage ('Try to login to installed cluster') {
+                        println "Waiting for successful login to the ${CLUSTER_NAME} cluster"
+                        i = 0
+                        // wait max 1 hour
+                        while(!loggedIn && i < 360) {
```
trepel: I would change the comment saying that this is for a specific number of tries (it is not really timeboxed). Eg `// Trying to log into the cluster (twelve tries should take approx. 1 hour)`. Also, if your aim is to have this waiting for roughly one hour, you need to decrease the number of tries to 12.

trepel: For loop might be more appropriate here.

jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
```diff
@@ -0,0 +1,206 @@
...
+                            loggedIn = ocLogin()
+                            i += 1
+                            sleep time: 5, unit: 'MINUTES'
+                        }
+                    }
+
+                    // we want to run test only when there was a successful login to the cluster
+                    if (loggedIn) {
```
trepel: what about having else brach making the build unstable (or failed might be even better). And outputting to log that the tests were skipped?

jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
```diff
@@ -0,0 +1,206 @@
...
+                                string(name: 'threeScaleBucketRegion', value: "${THREESCALE_BUCKET_REGION}"),
+                                string(name: 'backupAwsKey', value: "${AWS_ACCESS_KEY_ID}"),
+                                string(name: 'backupAwsSecret', value: "${AWS_SECRET_ACCESS_KEY}"),
+                                string(name: 'backupsBucketName', value: "${BACKUPS_BUCKET_NAME}"),
+                                string(name: 'clientSecret', value: "${SSO_SECRET}")
+                            ]).result
+                        }
+                    } // stage
+
+                    if (installationStatus != 'SUCCESS' || uninstallationStatus != 'SUCCESS') {
```
trepel: Does it make sense to proceed with the installation if uninstall fails? Why not to just fail right after the uninstall failure?

tremes: Well it probably doesn't. I am always not sure how the try/catch works in such case. I will split up this if - one after uninstallation and second after - installation

jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
tremes: yep good point

jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
tremes: Hmmm I can update the comment but I don't like using break in for loops TBH.

jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
```diff
@@ -0,0 +1,216 @@
...
+            try {
+                String loginOutput = sh (
+                    script: "oc login https://api.${CLUSTER_NAME}.openshift.com -u customer-admin -p ${CUSTOMER_ADMIN_PASSWORD}",
+                    returnStdout: true
+                ).trim()
+    
+                println loginOutput
+                return loginOutput.contains("Login successful")
+            } catch (error) {
+                println "Login error was " error
```
trepel: Are you sure this is a correct syntax? See https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/OSD/job/osd-testing-nightly-trigger/24/console

tremes: Yes it was wrong. Fixed. Thanks

jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
```diff
@@ -0,0 +1,217 @@
...
+                                booleanParam(name: 'CLEAN_RESOURCES', value:  Boolean.valueOf("false"))
+                            ])
+                        } // stage
+
+                        stage ('Create test user in SSO') {
+                            build(job: 'create-rhmi-user', propagate: false, parameters: [
+                                string(name: 'REPOSITORY', value: "${TEST_SUITES_REPOSITORY}"),
+                                string(name: 'BRANCH', value: "${TEST_SUITES_BRANCH}"),
+                                string(name: 'CLUSTER_URL', value: "https://console.${CLUSTER_NAME}.openshift.com"),
+                                string(name: 'ADMIN_USERNAME', value: "${ADMIN_USERNAME}"),
```
trepel: `01` missing here :) (I know you know)

trepel: actually no, it is defined as admin01 above

tremes: It's not AFAIK - https://github.com/integr8ly/ci-cd/pull/307/files#diff-5450a374ad8dd23c37ee56c49958b893R30

jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
```diff
@@ -0,0 +1,217 @@
...
+            GH_CLIENT_ID=836affcd7fd1bcebf9ae
+            GH_CLIENT_SECRET=72e97db6f1a2a6d8d2be66686cf52ccf15b3588f
+            CUSTOMER_ADMIN_PASSWORD=Password1
+            CREDENTIALS_ID=osd-s3-access-key-and-secret
+            THREESCALE_BUCKET_NAME=integreatly-qe1-3scale
+            THREESCALE_BUCKET_REGION=eu-west-1
+            BACKUPS_BUCKET_NAME=rhmi-qe1-backups
+            ADMIN_USERNAME=admin01
+            ADMIN_PASSWORD=Password1
+            CLEAN_RESOURCES=true
```
trepel: would it make sense to have this set to 'false', so one can check during the day?

jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
```diff
@@ -0,0 +1,217 @@
...
+                        }
+                    } // stage
+
+                    if (installationStatus != 'SUCCESS') {
+                        println 'There were some errors in previous installation pipeline!'
+                        currentBuild.result = 'FAILURE'
+                        return
+                    }
+
+                    // This relies on a cronjob (or something else) to update the client secret in master-config.yaml so that we can successfully login afterwards
```
trepel: Update comment, we know now that this relies on cronjob to restart api, not to fix client secret

jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
```diff
@@ -0,0 +1,217 @@
...
+            String body = "${subject}\n\n(${env.BUILD_URL})"
+
+            if(err != null) {
+                body = "${body}\n\n FAILURE was ${err}"
+            }
+
+            mail body: body, subject: subject, to: mailRecipients
+        }
+
+        def ocLogin() {
```
trepel: Very minor, but change name to `ocLoginAsCustomerAdmin()` since username is hard-coded there.

PR #310, opened 2019-10-14
removed patching grafana operator

pawelpaszki:
## What
removed patching grafana operator

## Why
to make sure that the monitoring works correctly (patching grafana operator can break the monitoring)

## Verification
Successful installation: https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/ppaszki-pds-install-ocp3/1/consoleFull


trepel:
To be honest I can't see the reason why Application Monitoring Operator and Grafana Operator should not work together if using 'master' branch for both. I understand that 'master' is unstable but it if does not work due to some large feature being developed spanning across multiple repositories, it should take days, not weeks to make our internal components work together again.

So I think we should keep testing 'master of everything' together. For sure, we can stop doing that altogether saying that we start doing that once our test suites are stable enough for 'master of installer'. But that is another story.

pawelpaszki:
@trepel closing this one (based on yesterday's conversation)

PR #313, opened 2019-10-17
INTLY-3858 - run nexus test only for PDS cluster

tremes:

### DIFF
--- a/jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
+++ b/jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
@@ -7,7 +7,7 @@
     sandbox: false
     concurrent: false
     triggers:
-      - timed: "H(20) 3 * * *"
+      - timed: "20 3 * * *"
     properties:
       - build-discarder:
           num-to-keep: 56
--- a/jobs/integr8ly/ocp3/test-suites/all-tests-executor.yaml
+++ b/jobs/integr8ly/ocp3/test-suites/all-tests-executor.yaml
@@ -143,9 +143,12 @@
                         runTests('sso-user-create-tests')
                     } // stage
 
-                    stage ('Nexus builds test') {
-                        runTests('nexus-builds-test')
-                    } // stage
+                    // run nexus-build test only on PDS
+                    if (CLUSTER_URL.contains("open.redhat.com")) {
+                        stage ('Nexus builds test') {
+                            runTests('nexus-builds-test')
+                        } // stage
+                    }
 
                     stage ('Start Dashboards Browser Tests') {
                         runSingleBrowserTest('tests/10_fuse/fuse-console-login.js')


tremes:
https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/all-tests-executor-tremes/1

jobs/integr8ly/ocp3/nightly/osd-testing-nightly-trigger.yaml
```diff
@@ -7,7 +7,7 @@
     sandbox: false
     concurrent: false
     triggers:
-      - timed: "H(20) 3 * * *"
+      - timed: "20 3 * * *"
```
tremes: I also added fix for time definition in the new osd nightly pipeline since the original is wrong.

jobs/integr8ly/ocp3/test-suites/all-tests-executor.yaml
```diff
@@ -143,9 +143,12 @@
                         runTests('sso-user-create-tests')
                     } // stage
 
-                    stage ('Nexus builds test') {
-                        runTests('nexus-builds-test')
-                    } // stage
+                    // run nexus-build test only on PDS
+                    if (CLUSTER_URL.contains("open.redhat.com")) {
```
trepel: Is it only me who thinks that this should be handled in the test level, not pipeline level. E.g. if there is no Nexus namespace the test file should be executed but the tests inside it should be programmatically skipped.

I think it is better to have our tests standalone as much as possible and pipeline should just trigger them.

tremes: TBH I am not sure how to handle it nicely in the test source code so if you have some suggestion then I am one ear.

trepel: what about this:
https://mochajs.org/#pending-tests (search for `this.skip()`) ?

trepel: It seems if you do that in beforeAll you can skip all the tests (which is what we want if Nexus namespace is not present.

tremes: OK interesting. This could be a cleaner solution (although I am not sure why the test is marked as "pending"), but on the other hand the stage will be still present in the pipeline, which could be a little bit confusing...but I don't have a strong preference here.

Feel free to close it and assign the issue to yourself :)

trepel: yeah, I really think it is better solution to do such things on the test level, not pipeline level since tests should be standalone and solve such things themselves.

I also think having the stage always there is less confusing than some stages only being present some times. Jenkins does not handle such situation nicely when displaying history and I think it is better to see the stage always there, so you can have a look and see that test were executed but effectively skipped - with a message why were skipped in the test log

trepel: Feel free to unassign yourself from the ticket fist ;)

PR #315, opened 2019-10-18
OSD nightly - fix time expression

tremes:


tremes:
I forgot about that yesterday :-(

PR #318, opened 2019-10-22
INTLY-3826 - update wt-* test executor definitions

tremes:


tremes:
Example is https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/all/job/w1-test-executor-tremes/2/

PR #319, opened 2019-10-22
Branch support for recreate pipelines

trepel:
## What

Added branch support for recreate-pipeline job. Successful build:
https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/Nightly/job/recreate-pipelines/273/


trepel:
I also moved the schedule a bit earlier (19:49 Brno time, 18:49 Irinsh). This needs to be triggered before all the other nightly pipelines which was not true.

PR #325, opened 2019-10-25
Parameterize OSD pipelines with tower instance name

tremes:
There is the option to have multiple tower instances configured in Jenkins. We temporarily need to have 2 instances connected/configured. That's why we parameterize that. 

`Dev Tower` is "old" common cicd instance.
`Temp Tower` is the newer one used only for 1.5.2 testing.


tremes:
It looks like it works https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/OSD/job/osd-cluster-integreatly-uninstall/75

tremes:
Pipelines recreated manually.

PR #327, opened 2019-10-31
Renaming Cluster Management Workflow Jobs

pmccarthy:
## What
Renaming Cluster Management jobs to reflect latest Tower Configurations

## Changes
Renamed `Provision Cluster` to `Cluster Provision`
Renamed `Deprovision Cluster` to `Cluster Deprovision`


mikenairn:
@pmccarthy Are we changing the tower url and credentials or are they going to be the same?

This looks fine anyway, will let you decide when it should be merged an applied to the QE jenkins.

pmccarthy:
> @pmccarthy Are we changing the tower url and credentials or are they going to be the same?
> 
> This looks fine anyway, will let you decide when it should be merged an applied to the QE jenkins.

@mikenairn No, the tower route and credentials will remain the same to avoid confusion. I'll be switching over the route to the new tower instance once everything gets merged back to master.

PR #332, opened 2019-11-01
Removed patching of Grafana operator

trepel:
## What

We cannot use the master branch of Grafana operator anymore since this branch contains pieces for OCPv4 now. After a lengthy conversation with DavidM it seems this is indeed the only easy and simple solution. The latest Grafana operator version for OCPv3 is v1.3.1.

Not sure what happens if some other modifications for OCPv3 are required. Probably branch from v1.3.1 will be created for that. We'll see I guess. I only commented the bit out since I still hope we get to pathing again at some point.


trepel:
@pawelpaszki pinging just you because you initiated this and you already created a https://github.com/integr8ly/ci-cd/pull/310 which I forced you to close. Sorry.

trepel:
@tonyxrmdavidson just fyi, this should solve the issue with Alerts we are seeing in pds 'master of everything' (a.k.a patched) pipeline.

PR #356, opened 2019-11-21
🔥 Remove MDC & MSS

grdryn:
JIRA:
https://issues.jboss.org/browse/AEROGEAR-10077
https://issues.jboss.org/browse/AEROGEAR-10078

Not sure what the state of "mobile-walkthrough" is, so I've left that. I think it will still exist but will be changed to not rely on either of these defunct services.


jobs/delorean/folders.yaml
```diff
@@ -159,39 +148,6 @@
...
-    name: delorean-mss-operator
-    display-name: 'MSS Operator'
-    project-type: folder
-
-- job:
-    name: delorean-mss-operator/ga
-    display-name: 'GA'
-    project-type: folder
-
 # UPS
```
grdryn: There's both a "UPS" and "UPS Operator" here. I think we spoke about this duplication before.

I can remove "UPS" and just have the "UPS Operator" section, if there's consensus for that?

mikenairn: Yes, we only need one of these jobs. We might as well just do it now. Ideally i think you would be better to delete the current ups one entirely and rename ups-operator to ups, and change all references to ups-operator, to ups. Reason being, it should match the name of the role in the installer https://github.com/integr8ly/installation/tree/master/roles/ups

b1zzu: Hey guys, I still have an open PR for removing the operand jobs #342 , but it's already outdated with the master, we can close this PR for removing MSS and MDC as it is and then I will rebase my PR in order to remove UPS and keep only the UPS operator, wdyt?

grdryn: @b1zzu this PR is now merged, so I'm not sure what to do about your PR. Is there stuff that yours does that's still valid?

PR #362, opened 2019-12-02
changed map to array

pawelpaszki:
## What
changed map to array due to Jenkins giving out about using permissions 

## Verification
Running here (with temporarily pasted pipeline definition instead of using script from GH: https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/OS4/job/openshift4-rhmi-image-deploy-install/5/console


tonyxrmdavidson:
LGTM

PR #363, opened 2019-12-02
added mechanism to not fail in the executor, if cluster with given na…

pawelpaszki:
…me already exists

## What
added mechanism to not fail in the executor, if cluster with given name already exists


tonyxrmdavidson:
LGTM

PR #378, opened 2020-01-06
CRDs and ApiServices removal in osd uninstall

trepel:
## What

Added removal of leftover CRDs and ApiServices to OSD uninstall job.


trepel:
The change has been applied (will be erased once recreate-pipelines is triggered) but not tested. If you happen to do uninstall on OSD, please review this alongside.

PR #383, opened 2020-01-16
fix: msb test - wait for pod to be ready

jhellar:
## What

MSB test pipeline sometimes fails [1] with:

```
--- FAIL: TestManagedBroker (0.21s)
01:39:00     managedServicesBroker_test.go:80: Error getting Catalog: 503 Service Unavailable
```

This is probably caused by pod not being ready yet. This change makes sure, that pod is ready before running the test.

[1] https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/msb-integration-test/319/console


trepel:
@jhellar hmm, it seems it does not solve the issue

jhellar:
Rerun the test with https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/msb-integration-test/321/replay/diff, seems that this might help.

PR #393, opened 2020-01-27
EWS discovery job updates

mikenairn:
## What

* Only apply product patch updates to master if we have opted to not update the latest release for this product.
* Re-enabled rhsso ews jobs
* Set updateLatestRHMIRelease to true for rhsso, 3scale, amq-online and fuse-online. All others will just PR any updates, including patch updates, back to master.

### DIFF
--- a/jobs/delorean/1.0/ews/codeready/ga/discovery.yaml
+++ b/jobs/delorean/1.0/ews/codeready/ga/discovery.yaml
@@ -40,7 +40,7 @@
           read-only: true
       - bool:
           name: 'updateLatestRHMIRelease'
-          default: false
+          default: true
           description: '[OPTIONAL] Update the latest RHMI release branch with patch releases!'
     pipeline-scm:
       script-path: jobs/delorean/jenkinsfiles/1.0/ews/discovery/ga/github/Jenkinsfile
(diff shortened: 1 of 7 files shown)


mikenairn:
The point of this change is to try and get the patch updates for products back to the appropriate branch while ensuring we don't add upgrade logic in 2 places unnecessarily.

rhsso, 3scale, amq-online and fuse-online will now only apply updates for patch releases to the latest released version of integreatly (v1.6 currently). Major and Minor version updates of these will continue to be applied to master.

It will be up to the person dealing with the patch release PR e.g. https://github.com/integr8ly/installation/pull/1249 to ensure that the manifest update https://github.com/integr8ly/installation/pull/1249/commits/f323b7a5b334333e7ce83bdd2693ccf2300c8a59 gets rolled back into master without the upgrade part https://github.com/integr8ly/installation/pull/1249/commits/67fb43974055bfc8713f65bfc2692857124cb3db

jobs/delorean/1.0/ews/codeready/ga/discovery.yaml
```diff
@@ -40,7 +40,7 @@
           read-only: true
       - bool:
           name: 'updateLatestRHMIRelease'
-          default: false
+          default: true
```
mikenairn: @pb82 codeready is going to work the same way as the rest.

PR #411, opened 2020-02-12
fix: switch off use of CRO resources temporarily

damienomurchu:
## What
https://issues.redhat.com/browse/INTLY-5315
Switch off the use of CRO resources and use in-cluster resources instead.

## Why
OSD nightly pipeline currently failing as test clusters provisioned in OSD AWS account, and we don't have permissions to create the required CRO resources in that account. Until this issue is resolved this change is to switch over to using in-cluster resources instead.

As long as we use CRO resources in the install, the deployment of 3scale, CodeReady, and UPS will not complete. Pipeline: https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/openshift4-rhmi-image-deploy-install/103/console

## How 
switch off use of CRO resources by updating CRO flag in pipeline

### DIFF
--- a/jobs/integr8ly/ocp4/install/image-deploy/Jenkinsfile
+++ b/jobs/integr8ly/ocp4/install/image-deploy/Jenkinsfile
@@ -14,7 +14,7 @@ String installationNamespace = "redhat-rhmi-operator"
 
 String expectedOcCommandOutput = ""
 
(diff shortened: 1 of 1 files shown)


pawelpaszki:
Looks good to me. Did you have a successful run with the changed flag?

PR #412, opened 2020-02-13
fix: add missing port to cluster api uri

damienomurchu:
https://issues.redhat.com/browse/INTLY-5333


pawelpaszki:
@damienomurchu I don't think it is going to work against OS3 clusters

PR #419, opened 2020-02-17
Intly 5341 product branch management stages

StevenTobin:
## What
Adds branch management stages for the ews

* Checkout operator repo
* Create the product_name branch if it doesn't exist
* Check for changes against the remote and push if there are any detected

### DIFF
--- a/jobs/delorean/2.0/ews/3scale/discovery.yaml
+++ b/jobs/delorean/2.0/ews/3scale/discovery.yaml
@@ -22,6 +22,14 @@
     - string:
           name: CI_MESSAGE
           description: This will be populated by the CI trigger
+    - string:
+          name: productBranch
+          description: The name of the product branch
+          default: '3scale-next'
+    - string:
+          name: baseBranch
+          description: The branch to base the product-next branch off
+          default: 'delorean-ews'
 
     pipeline-scm:
       script-path: jobs/delorean/jenkinsfiles/2.0/ews/discovery/umb/Jenkinsfile
--- a/jobs/delorean/jenkinsfiles/2.0/ews/discovery/umb/Jenkinsfile
+++ b/jobs/delorean/jenkinsfiles/2.0/ews/discovery/umb/Jenkinsfile
@@ -1,9 +1,11 @@
 pipeline {
-    agent none
+    agent any
 
     parameters {
         // https://datagrepper.engineering.redhat.com/id?id=ID:messaging-devops-broker02.web.prod.ext.phx2.redhat.com-44775-1580379079663-8:893046:0:0:1&is_raw=true&size=extra-large
         string defaultValue: '', description: 'This will be populated by the CI trigger', name: 'CI_MESSAGE', trim: true
+        string defaultValue: '', description: 'The name of the product branch', name: 'productBranch'
+        string defaultValue: 'delorean-ews', description: 'The base branch for the <product>-next branch', name: 'baseBranch'
     }
 
     stages {
@@ -17,15 +19,30 @@ pipeline {
 
         stage('Clone Operator Repo') {
             steps {
-                //ToDo Clone the integeratly-operator repo
-                echo "Clone Operator Repo"
+                cleanWs()
+                git branch: '${baseBranch}', url: 'https://github.com/integr8ly/integreatly-operator'
             }
         }
 
         stage("Ensure Product Branch") {
             steps {
-                //ToDo Create the product next branch i.e. 3scale-next from the target branch if it doesn't exist
-                echo "Ensure Product Branch"
+                withCredentials([usernamePassword(credentialsId: 'githubjenkins', passwordVariable: 'gitPassword', usernameVariable: 'gitUsername')]) {
+                    sh("""
+                        git config --global user.name 'Automatron'
+                        git config --global user.email 'github.eng@feedhenry.com'
+                        git config --global credential.username ${gitUsername}
+                        git config --global credential.helper "!echo password=${gitPassword}; echo"
+                       """)
+                }
+                script {
+                    def checkout = sh(script: 'git checkout ${productBranch}', returnStatus: true)
+                    if (checkout) {
+                        sh("""
+                            git checkout -B ${productBranch}
+                           """)
+                    }
+                    sh 'git rebase ${baseBranch}'
+                }
             }
         }
 
@@ -52,8 +69,12 @@ pipeline {
 
         stage('Push Branch') {
             steps {
-                //ToDo Push the product branch if an changes have been made
-                echo "Push Branch"
+                script {
+                    def changes = sh(script: 'git log ${productBranch}..origin/${baseBranch} --pretty=o | wc -l', returnStdout: true)
+                    if ((changes as int) > 0) {
+                        sh 'git push origin ${productBranch} --force'
+                    }
+                }
             }
         }
 
--- a/jobs/delorean/2.0/ews/amq-online/discovery.yaml
+++ b/jobs/delorean/2.0/ews/amq-online/discovery.yaml
@@ -22,6 +22,14 @@
     - string:
           name: CI_MESSAGE
           description: This will be populated by the CI trigger
+    - string:
+        name: productBranch
+        description: The name of the product branch
+        default: 'amq-online-next'
+    - string:
+        name: baseBranch
+        description: The branch to base the product-next branch off
+        default: 'delorean-ews'
 
     pipeline-scm:
       script-path: jobs/delorean/jenkinsfiles/2.0/ews/discovery/umb/Jenkinsfile
--- a/jobs/delorean/2.0/ews/fuse-online/discovery.yaml
+++ b/jobs/delorean/2.0/ews/fuse-online/discovery.yaml
@@ -22,6 +22,14 @@
     - string:
           name: CI_MESSAGE
           description: This will be populated by the CI trigger
+    - string:
+        name: productBranch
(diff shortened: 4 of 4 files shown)


jobs/delorean/2.0/ews/3scale/discovery.yaml
```diff
@@ -22,6 +22,9 @@
     - string:
           name: CI_MESSAGE
           description: This will be populated by the CI trigger
+    - string:
+          name: PRODUCT_BRANCH
+          description: The name of the product branch
```
laurafitzgerald: How does this one get populated?

jobs/delorean/jenkinsfiles/2.0/ews/discovery/umb/Jenkinsfile
```diff
@@ -17,15 +17,31 @@ pipeline {
 
         stage('Clone Operator Repo') {
             steps {
-                //ToDo Clone the integeratly-operator repo
-                echo "Clone Operator Repo"
+                cleanWs()
```
StevenTobin: This should prob be higher up actually so im not wiping anything that happens earlier

jobs/delorean/jenkinsfiles/2.0/ews/discovery/umb/Jenkinsfile
```diff
@@ -17,15 +17,31 @@ pipeline {
...
+                        git config --global credential.username ${GIT_USERNAME}
+                        git config --global credential.helper "!echo password=${GIT_PASSWORD}; echo"
+                       """)
+                }
+                script {
+                    def checkout = sh(script: 'git checkout ${PRODUCT_BRANCH}', returnStatus: true)
+                    if (checkout) {
+                        sh("""
+                            git checkout -B ${PRODUCT_BRANCH}
+                            git push origin ${PRODUCT_BRANCH}
```
laurafitzgerald: Looks like there's a push in the later stage so maybe it's not needed here?

jobs/delorean/jenkinsfiles/2.0/ews/discovery/umb/Jenkinsfile
```diff
@@ -17,15 +17,31 @@ pipeline {
 
         stage('Clone Operator Repo') {
             steps {
-                //ToDo Clone the integeratly-operator repo
-                echo "Clone Operator Repo"
+                cleanWs()
+                git 'https://github.com/integr8ly/integreatly-operator'
```
laurafitzgerald: This is nice. Is there more ways to use this other than cloning? I didn't like using so much `sh` everywhere but didn't know that the clone could be done like this.

jobs/delorean/jenkinsfiles/2.0/ews/discovery/umb/Jenkinsfile
```diff
@@ -52,8 +68,12 @@ pipeline {
 
         stage('Push Branch') {
             steps {
-                //ToDo Push the product branch if an changes have been made
-                echo "Push Branch"
+                script {
+                    def changes = sh(script: 'git log origin/${PRODUCT_BRANCH}..HEAD --pretty=o | wc -l', returnStdout: true)
+                    if ((changes as int) > 0) {
+                        sh 'git push origin ${PRODUCT_BRANCH} --force'
```
laurafitzgerald: Would there need to be a commit in here before the push?

jobs/delorean/2.0/ews/3scale/discovery.yaml
StevenTobin: Ah yeah, forgot to default it. I'll add that in

jobs/delorean/jenkinsfiles/2.0/ews/discovery/umb/Jenkinsfile
StevenTobin: So this is to catch the case where the branch does not exist. I think pushing here is fine because if there's an intermediate step added that reaches out to the remote we'll need the remote to exist.

jobs/delorean/jenkinsfiles/2.0/ews/discovery/umb/Jenkinsfile
StevenTobin: I'm not sure really, should we commit here or in the process step where we know what changes need to be made? maybe revisit when it's more fleshed out?

jobs/delorean/jenkinsfiles/2.0/ews/discovery/umb/Jenkinsfile
mikenairn: We need to be able to specify the branch here. Will eventually be 'master', but should be set to "deloeran-ews" for the time being.

The stages after this will also need access to the base branch so we need to deal with that also.

jobs/delorean/jenkinsfiles/2.0/ews/discovery/umb/Jenkinsfile
```diff
@@ -17,15 +17,31 @@ pipeline {
...
+                }
+                script {
+                    def checkout = sh(script: 'git checkout ${PRODUCT_BRANCH}', returnStatus: true)
+                    if (checkout) {
+                        sh("""
+                            git checkout -B ${PRODUCT_BRANCH}
+                            git push origin ${PRODUCT_BRANCH}
+                           """)
+                    }
+                    sh 'git rebase master'
```
mikenairn: This needs to be configurable. We should be able to set the target branch somewhere and have it be used everywhere.

jobs/delorean/jenkinsfiles/2.0/ews/discovery/umb/Jenkinsfile
mikenairn: Does this clone the repo and change the directory into it?

jobs/delorean/jenkinsfiles/2.0/ews/discovery/umb/Jenkinsfile
mikenairn: Each stage will deal with committing changes that it makes. This stage should only push the branch if there are changes to push.

jobs/delorean/jenkinsfiles/2.0/ews/discovery/umb/Jenkinsfile
mikenairn: We had this setup originally with DeLorean 1.0 but the number of branches that were hanging around with no changes on them was confusing.

The current approach is the create the branch here if one doesn't already exist, or checkout the existing one, and not to push the branch until the end if changes are detected between it and the base branch (master)

jobs/delorean/jenkinsfiles/2.0/ews/discovery/umb/Jenkinsfile
mikenairn: It needs to one of the first things that you do alright, but i think here is fine tbh. There should be nothing happening before this that is doing anything in the workspace anyway.

jobs/delorean/jenkinsfiles/2.0/ews/discovery/umb/Jenkinsfile
StevenTobin: It does clone and change into the newly cloned directory.

I don't know if you can specify a branch here, I think we'd have to `sh git checkout` but I can add that so we start at the branch we want, it defaults to master.

jobs/delorean/jenkinsfiles/2.0/ews/discovery/umb/Jenkinsfile
StevenTobin: I'll add it as a parameter to the job.

jobs/delorean/jenkinsfiles/2.0/ews/discovery/umb/Jenkinsfile
StevenTobin: Ignore me, it's really easy, change pushed

jobs/delorean/2.0/ews/3scale/discovery.yaml
```diff
@@ -22,6 +22,14 @@
     - string:
           name: CI_MESSAGE
```
mikenairn: @psturc Are you using uppercase variable names in the new jenkins configuration repo? Is that something that was decided upon?

psturc: @mikenairn in the end we decided to go with camelCase: https://gitlab.cee.redhat.com/integreatly-qe/ci-cd/blob/master/jobs/Integreatly/rhmi-install-addon-flow.groovy

jobs/delorean/jenkinsfiles/2.0/ews/discovery/umb/Jenkinsfile
StevenTobin: I'm not sure really, here's the docs https://jenkins.io/doc/pipeline/steps/git/

mikenairn:
> Is believe the plan is to use <product>-next for the branch name. Is there any advantage to switching to a new one based on product and version @mikenairn ?

@laurafitzgerald For the first iteration of this we are keeping it fairly simple and just going with a single "next" branch per product that will be updated with the latest thing we had discovered and target going back to master for everything.

As we progress the work more i can image we might change how this works, and having a branch dedicated to a particular release might be the way we want to go.

mikenairn:
Verified this is working as expected. Was able to pull and push a branch(with small mod to force it).

PR #450, opened 2020-03-23
Del 202 move ews scripts

mikenairn:
## What
*  Update umb discovery job to checkout delorean repo and use scripts in that repo.
* Update discovery jobs to use master as the default sourceBranch.
* Dont send email if dryrun is true.
* Update dry run of jira and pr stages to set url variable values when appropriate.


StevenTobin:
:eyes:

PR #457, opened 2020-04-03
update AWS details and host details for new Eng tower

damienomurchu:


damienomurchu:
@StevenTobin will use these to rebuild jobs on our Jenkins and then run some test jobs to verify

damienomurchu:
@StevenTobin closing this temporarily. If we need to re-open it we can.

PR #472, opened 2021-02-24
remove redundant all-tests-executor stages

damienomurchu:
## What

Remove tests for unsupported products from pipeline: [JIRA](https://issues.redhat.com/browse/INTLY-10319)

## Why

Only AMQ, 3Scale, and SSO are now supported in RHMI, so redundant tests for non-supported components should be removed so they do not have to be maintained or run during release testing.

## How

Remove stages in all-tests-executor pipeline that runs tests related to CodeReady and Fuse.


psturc (approved):

damienomurchu:
Thanks for review @psturc :+1:

